PR #1, opened 2015-08-11
WebSocket support for gevent and more

drdaeman:
Hello,

First of all, thanks for your project! I was somehow frustrated by lack of modern and maintained KISS bidirectional transport libraries, until I found your on PyPI.

For some reason eventlet's websocket implementation had only worked in Chromium, but not Firefox. So I thought rather than trying to debug eventlet's implementation, I'd use gevent - which is my favorite anyway - and surely I still wanted have WebSockets transport ;)

The overall idea of this pull request is simple. To abstract between various async libraries I've replaced `Server.async` dict with pluggable modules from `engineio.async` that have fairly uniform API.  Then I wrote a simple wrapper for gevent-websocket to resemble eventlet.websocket's conventions.

While I was at it, I've also fixed two unrelated minor bugs. First, I've made sure `pkt.packet_type` is validated in `Socket.receive`. Then, I've improved parsing of "Connection" header in `handle_get_request` a bit. Sorry for a mostly monolithic commit - the modifications evolved after some trial-and-error and intermediate results were a terrible mess I didn't dare to commit.

Hope you'll find this useful. Sorry if not.


engineio/async/eventlet.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
```
miguelgrinberg: Is the __future__ thing really needed? This project does not run on anything below 2.7.

engineio/async/gevent.py
```diff
@@ -0,0 +1,51 @@
...
+
+def wrap_websocket(app):
+    """
+    Given a callable that accepts a single argument - the websocket object,
+    make gevent-websocket's approach compatible with eventlet's.
+    """
+    def __wrapper(environ, start_response):
+        ws = WebSocket(environ["wsgi.websocket"])
+        return app(ws)
+    return __wrapper
```
miguelgrinberg: I'm going to test this carefully, but it appears this approach works with the standard wsgi server instead of the custom server that comes with gevent-websocket, right?

If this is correct, then it is fantastic. One of the reasons I went with eventlet is the simplicity in which the socket connection is handled.

engineio/async/threading.py
```diff
@@ -0,0 +1,19 @@
...
+QueueEmpty = queue.Empty
+has_websocket = False
+
+
+def thread(target):
+    return threading.Thread(target=target)
+
+
+def wrap_websocket(app):
+    raise RuntimeError("Websocket support is not available")
```
miguelgrinberg: I think it may be a bit more clear to make `wrap_websocket()` return None to indicate this transport isn't available. That way you don't need to have `has_websocket`.

miguelgrinberg:
Wow. This is a pretty awesome contribution. Thanks for taking the time to do it!

There are a few things I'd like to reorganize to get this change ready to merge. Please let me know how much of this work you are interested in doing (I'll handle the rest).

1. I would like to have separate PRs for each logical change you've made so that I can merge them separately. I think we need one for the executable bits, one for the `packet_type`  check, one for the `strip()` in the connection parsing, and finally the big one for the async improvements.

2. The commits for the async improvements should be squashed into a single commit, so that it appears as a single entry once I do the merge (I use the commit logs as change logs, single commit merges help keeping the log clean and clear).

3. The test coverage dropped to 98%, I try to keep it at 100%. I'm guessing some of the new code you wrote needs specific test(s).

Let me know if you want to do any of this, or if you prefer that I handle it myself.

Thanks again!

miguelgrinberg:
Oh, and as a side note, Firefox does work for me with eventlet's WebSockets. Odd.

engineio/async/eventlet.py
drdaeman: I believe, yes, because 2.7 still has relative ones by default. Without this test project fails with ImportErrors for me, trying to import non-existent `engineio.async.gevent.queue` when it sees `import gevent.queue`.

engineio/async/gevent.py
drdaeman: Sad to disappoint, but no. This only works with either `geventwebsocket.WebSocketServer` as a server or if `handler_class=WebSocketHandler` is passed to `gevent.pywsgi.WSGIServer` constructor.

As far as I know, there seems to be [no common standard](https://groups.google.com/forum/#!topic/gevent/JPRoBdeLTbE) on passing the socket handle to a WSGI application, so every library out there does it differently. So for each server one would need some magic trick, like [uwsgi-websockets](https://github.com/gavinwahl/uwsgi-websockets).

Maybe my understanding is wrong or outdated, though.

engineio/async/threading.py
drdaeman: I went with additional boolean flag because `Server._upgrades` wants to know whenever there is WS support available or not. I could've also made it that `wrap_websocket(None)` would be a convention to check if the support is available without actually wrapping anything, but boolean variable looked simpler.

drdaeman:
Thanks. Sure, I'll split the PR and add extra test cases to bring back 100% coverage for `server.py` and `socket.py`. Not sure about today, more likely tomorrow, though.

But I'm not sure how to test `async/*` modules, which currently have awful 6-8% coverage. Writing tests for actual event libraries (likely, starting a server and doing a full round-trip of messages) sounds good, but would likely require much more time than I can spend. And there isn't much else to test in those modules - checking things like that `thread(f)` would really call `gevent.Greenlet(f)` doesn't make much sense to me.

engineio/async/gevent.py
miguelgrinberg: The reason I ask is that having to specify a custom gevent server is source of confusing for many users. I actually don't mind a magic trick, as long as it's all done from our side without the user having to run in a special configuration. I actually didn't know uwsgi-websockets, that sounds very interesting, a similar solution might work for this.

engineio/async/eventlet.py
```diff
@@ -0,0 +1,20 @@
...
+    has_websocket = True
+except ImportError:
+    has_websocket = False
+
+Queue = eventlet.queue.Queue
+QueueEmpty = eventlet.queue.Empty
+
+
+def thread(target):
+    return eventlet.green.threading(target=target)
```
miguelgrinberg: This should be `eventlet.green.threading.Thread(target=target)`.

engineio/async/eventlet.py
miguelgrinberg: I think there is some odd interaction with `importlib` that is causing these problems, because 2.7 supports absolute imports just fine. I'll try a few things to see why this happens, I think this should work without using the future import.

engineio/async/gevent.py
miguelgrinberg: I think sending the gevent traffic to the WebSocketHandler class from gevent-websocket should be possible in the middleware, without forcing users to use a non-standard server class. One more thing I need to research, but I'm think it can be done, which means that this server should run without having to do any tricks on uwsgi and gunicorn.

miguelgrinberg:
@drdaeman Hey, I'm going to start submitting commits derived from your PR, if you don't mind. Let me know if you were planning to work on this, but if not I think I can take it to the finish line myself. Thanks!

drdaeman:
@miguelgrinberg Uh... sorry, I promised I'd finish things, but unexpectedly got busy with some other matters and wasn't able to work on this. Should have commented earlier, but hoped I'd still find time... My bad, but it seems I'll leave it up to you. :(

The only thing I've did so far is figuring out that abovementioned uwsgi-websockets is broken and dead and doesn't work with recent gevent-websockets anymore. I've tried to sort this out, but ended up ditching uWSGI and running bare gevent's pywsgi-based servers.

Sorry.

miguelgrinberg:
@drdaeman No worries, I'm excited about this feature, happy to finish it myself.

Regarding uWSGI, I was thinking about it too. I think it would be possible to create another async module to use the native websocket support in offers (without relying on eventlet or gevent). There is a Python API for it, so I think it may be possible to use the same trick of creating a wrapper class that adapts the interface to look similar to eventlet.

miguelgrinberg:
@drdaeman I'm closing this PR, as I believe I have included all your changes, and you do have to use the websocket specific handler that comes with gevent-websocket, unfortunately. One of these days I'll try to write an alternative gevent module that uses the uwsgi websockets, I think that should be possible with this extensible design.

Thanks for all your help!

PR #4, opened 2015-08-29
Added b64 checks and encoding during initial connect

meismyles:
- Added b64 as passed parameter to '_handle_connect' and '_ok' and included in relevant calls.
- Added check to determine if b64 = '1' or 'true'. With just the above changes, b64 could = anything (including '0', 'false', etc.) and b64 encoding would still be applied.


miguelgrinberg:
Thanks!

PR #9, opened 2015-12-11
_websocket_handler waits on writer even after the socket is closed

babu-shanmugam:
This patch wakes up the writer with a null message when the socket gets closed. Without this the writer() waits unnecessarily which makes the serving thread to wait longer.


miguelgrinberg:
Yeah, this looks much better. Thanks!

PR #12, opened 2016-01-02
Fix eventlet wsgi websocket __call__ return

artemiy-rodionov:
This patch fixes websocket disconnect and resolves flask-socketio issue https://github.com/miguelgrinberg/Flask-SocketIO/issues/184. Eventlet WebSocketWSGI  returns a value - https://github.com/eventlet/eventlet/blob/master/eventlet/websocket.py#L135


miguelgrinberg:
Awesome, thanks a lot!

PR #15, opened 2016-02-23
Polling: allow GET requests until send queue is empty

fredericsureau:
When using polling transport, calling close() blocks GET requests so CLOSE packet can't be sent.

This pull request adds a writeable flag that block send() until the socket can be actually closed

### DIFF
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -18,6 +18,7 @@ class Socket(object):
         self.connected = False
         self.upgraded = False
         self.closed = False
+        self.writeable = True
 
     def poll(self):
         """Wait for packets to send to the client."""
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
You can call `close(wait=False)` to avoid blocking on a queue when the socket is going away. Doesn't that work for you?

fredericsureau:
I actually want to block on the queue. I want to wait for the CLOSE packet to be sent.
But GET requests are discarded because the `closed` flag is set.
In the polling transport, GET requests are the only way to send packets to the client.
So my PR ensures messages are flushed before setting socket as closed.

miguelgrinberg:
ah, sorry, I misunderstood. Now it makes more sense. Before I merge, I need to test it a bit. I'm not sure I understand why none of the unit tests needed to be adapted for this change, it looks like I do not have a test case that covers this particular usage. Thanks!

miguelgrinberg:
This is already implemented with the `Socket.closing` attribute.

PR #18, opened 2016-04-21
Starting thread in daemon mode

PierrePaul:
Starting thread in daemon mode, so it no longer blocks the main process when closing.
Like I said at [flask-socketio](https://github.com/miguelgrinberg/Flask-SocketIO/issues/237) , I'm not entirely sure it's a good idea, but all the async_mode I've seen use the std Thread class which supports the daemon attribute.

Here is some documentation about the attribute : https://docs.python.org/2/library/threading.html#threading.Thread.daemon
https://docs.python.org/3/library/threading.html#threading.Thread.daemon

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -281,7 +281,8 @@ class Server(object):
         """
         th = getattr(self.async['threading'],
                      self.async['thread_class'])(target=target, args=args,
-                                                 kwargs=kwargs, daemon=True)
+                                                 kwargs=kwargs)
+        th.daemon = True
         th.start()
         return th  # pragma: no cover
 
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Not sure what this change improves. This is a server, it's not supposed to exit unless it is killed with a signal.

PierrePaul:
Will try to make a demo repo tomorrow morning. It's mainly useful when
calling ```python manage.py db upgrade``` (it never dies ) or when running
tests.

miguelgrinberg:
you shouldn't be starting background threads when you are not running the server.

PR #21, opened 2016-05-13
Prevent raising an error so thread handling the redis won't die

cmabastar:
I added an issue in the other repo. This fix seems to have resolve it so far.

[https://github.com/miguelgrinberg/Flask-SocketIO/issues/263](url)

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -336,8 +336,8 @@ class Server(object):
         except KeyError:
             raise KeyError('Session not found')
         if s.closed:
-            del self.sockets[sid]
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Raising in the `_get_socket()` method makes sense. This is a low level function, if it is asked to return a socket for a client that went away it can't possibly do it. Your change removes the exception, which will cause the function to return a socket object that is stale.

This should be addressed higher in the call stack.

miguelgrinberg:
I'm going to close this PR, but please try the commit I just added on master. I think that will address your issue in a safer way.

PR #24, opened 2016-06-22
holdtime for  server to have a flexible timer when marking the client as close

cmabastar:
Dear Miguel,

This is an implementation of my ping_holdtime patterned after the BGP & LDP hello/keepalive holdtime.

Base on a series of tests on my redis/eventlet environment, with 
- very low ping_interval= 2
- high burst of traffic = .8 msg/per sec + another 1 msg/per sec
- and multiple connection 30+  ( I know 30 is very low )

I have found out that some clients are timing out due to a busy processing on the server end.  With the additional latency from redis and pub/sub processing overhead to send messages to multiple clients, the 25% leeway on the ping interval calculation might not be enough in certain use cases.

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -27,6 +27,8 @@ class Server(object):
                          server to respond before disconnecting.
     :param ping_interval: The interval in seconds at which the client pings
                           the server.
+    :param ping_holdtime: The time in which the server considers the client as 
+                          timeout. (Recommended 3 times of ping_interval)
     :param max_http_buffer_size: The maximum size of a message when using the
                                  polling transport.
     :param allow_upgrades: Whether to allow transport upgrades or not.
@@ -57,9 +59,12 @@ class Server(object):
                  max_http_buffer_size=100000000, allow_upgrades=True,
                  http_compression=True, compression_threshold=1024,
                  cookie='io', cors_allowed_origins=None,
-                 cors_credentials=True, logger=False, json=None, **kwargs):
+                 cors_credentials=True, logger=False, json=None, 
+                 ping_holdtime=75,
+                 **kwargs):
         self.ping_timeout = ping_timeout
         self.ping_interval = ping_interval
+        self.ping_holdtime = ping_holdtime
         self.max_http_buffer_size = max_http_buffer_size
         self.allow_upgrades = allow_upgrades
         self.http_compression = http_compression
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -55,7 +55,7 @@ class Socket(object):
         """Send a packet to the client."""
         if self.closed:
             raise IOError('Socket is closed')
-        if time.time() - self.last_ping > self.server.ping_interval * 5 / 4:
+        if time.time() - self.last_ping > self.server.ping_holdtime:
             self.server.logger.info('%s: Client is gone, closing socket',
                                     self.sid)
             self.close(wait=False, abort=True)
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -56,7 +56,7 @@ class Socket(object):
         if self.closed:
             raise IOError('Socket is closed')
         if time.time() - self.last_ping > self.server.ping_holdtime:
-            self.server.logger.info('%s: Client is gone, closing socket',
+            self.server.logger.info('%s: Client is beyond hold timer, closing socket',
                                     self.sid)
             self.close(wait=False, abort=True)
             return
--- a/tests/test_socket.py
+++ b/tests/test_socket.py
@@ -20,6 +20,7 @@ class TestSocket(unittest.TestCase):
         mock_server = mock.Mock()
         mock_server.ping_timeout = 0.2
         mock_server.ping_interval = 0.2
+        mock_server.ping_holdtime = 0.6
 
         try:
             import queue


miguelgrinberg:
@cmabastar ping interval is controlled by the client, not by the server. Shouldn't this be a change that goes in the socket.io clients?

I am working on changes that I think will address the timeouts. But in your application, this would be easy to address as well, just make sure all events return quickly. If you have a long-running task, just start a thread for it, but always let event handlers return quickly. The change that I'm currently testing basically runs all handlers in threads, so that you don't have to.

cmabastar:
Hi @miguelgrinberg 

Yes, ping_interval is handled by the clients, however the server checks the last_ping calculation and it offers a 25% leeway on it.  This fix is practically just to have a better config on the leeway.

```diff
-        if time.time() - self.last_ping > self.server.ping_interval * 5 / 4:
-            self.server.logger.info('%s: Client is gone, closing socket',
+        if time.time() - self.last_ping > self.server.ping_holdtime:
+            self.server.logger.info('%s: Client is beyond hold timer, closing socket',
```

Also, the eventhandlers I have is pretty fast and asynchronous, I don't even allow them to run emit directly, I call celery tasks to handle emission and the celery worker does the emit (goes via redis).  Maybe my app too I/O bound this way, that made packets backlog during a high emission testing.

miguelgrinberg:
Your solution is (I think) hacky, because it only looks at the problem from one side. If we make the assumption that the client isn't backed up (i.e. only the server is), then the client is sending the ping packets at the correct interval, and the responses to them are getting delayed. So even if you give the server extra time to respond with the pong packet, the client is in its perfect right to declare that the server is gone, since the pong packets aren't sent promptly. So the client will still drop and then reconnect. I'm sure you tested it on your particular set up and with the Socket.IO client that you are using, but as a general solution it does not seem viable, in my opinion.

christopher-abastar:
Hello miguel,

I think you were referring to the "ping_timeout" this is the value that as you have said makes the client terminates the server in which I have no issues with.  

The changes I did was more on the server perspective terminating a client which is the "ping_interval".

> If we make the assumption that the client isn't backed up (i.e. only the server is), then the client is sending the ping packets at the correct interval, and the responses to them are getting delayed.

Yes, you are right the client did send at the correct interval, however the server didn't even reach to the point to send a PONG response to the client because he already pass the "ping_interval". 

This scenario only applies if there's a backlog of messages in the packets. Let's say the following diagram where in d or data in the queue:

```ascii
d = data
ping = ping from client
t = time in seconds to n

           client ping at t(2)
            |      

       d1  d2   d3   ping
 t   --1----2----3----4------------------------------------------> n
```

In this example, the client sent the data correctly every 2 seconds. However, since there are so many backlog in the queue, server was able to process it at 4th second. This will make the server think the client has disconnected.

miguelgrinberg:
Frst of all, in your example you are talking about a 2 second ping interval. The default interval for Socket.IO is 25 seconds. The client sends a ping every 25s, then waits 60s for the response. It sounds like you are using values that are much much shorter than the defaults. Is there a reason for that?

Second, while I understand the backlog of messages situation that you are describing. And your change gives the server more time to respond without closing the connection due to a ping timeout. What I don't follow is how can you make the client wait more for a delayed pong response. If the client sends a ping at some time t, but the server only receives it at t+d (with d being some long delay), the only way you can prevent the client from closing the connection is by making the ping timeout larger than the maximum possible d. This is sort of accounted for in the 25s interval vs. the 60s timeout. If things are backed up in a way that makes the server unable to respond to a ping in less than 60s, then I think there are bigger problems in the application that need to be addressed.

christopher-abastar:
Hi Miguel,

> Frst of all, in your example you are talking about a 2 second ping interval. The default interval for Socket.IO is 25 seconds. The client sends a ping every 25s, then waits 60s for the response. It sounds like you are using values that are much much shorter than the defaults. Is there a reason for that?

Yes, currently we are testing the client's auto-reconnect feature and we had to set it on a low interval to simulate it. Also we had to put a low value to test server performance handle multiple client connections.  Client ping was set to 2s, but server response is still at it's default of 60s.

> What I don't follow is how can you make the client wait more for a delayed pong response. If the client sends a ping at some time t, but the server only receives it at t+d (with d being some long delay), the only way you can prevent the client from closing the connection is by making the ping timeout larger than the maximum possible d. 

I'm not sure if the commit changes that I did affected this behavior (Sending a delayed PONG response).  The only intention of the changes I made was to add a small leeway when the server disconnects a certain client.

Also to clarify, I think there were some missing points on my diagram earlier.
Given the same situation, let's say when d2 is being processed. This data is sent from a different client _**(client-x)**_ and after processing this data or event, it was required to send a broadcast/multicast to all connected clients. By the time the server tried to send this data to our _**client-1**_ at time=t(2) it already detected it exceeded "timeout_interval" thus disconnecting the **_client-1_**

```ascii
d = data
d2 = client-x sent this data and then had to forward to all clients
ping = ping from client
t = time in seconds to n

           client1 ping at t(2)
            |      

       d1  d2   d3   ping
 t   --1----2----3----4------------------------------------------> n
```

miguelgrinberg:
> Yes, currently we are testing the client's auto-reconnect feature and we had to set it on a low interval to simulate it.

Still not convinced this is right. Reconnections are an exceptional situation, that does not happen normally. The taxing on the system you are imposing by lowering from 25s to 2s is actually making it more likely that things will get backed up. I would actually go the other way, and allow pings to be sent less frequently. Maybe you don't realize that the client and the server do not rely on pings alone to detect disconnection. This is the method used when doing long-polling, but when the connection is over WebSocket, a disconnection is detected immediately, regardless how ping/ping packets.

Do ping packets get delayed when you use the default 25/60 interval and timeout?

> I'm not sure if the commit changes that I did affected this behavior (Sending a delayed PONG response)

Yeah, I believe it does. If you are saying that you want the hold time to be 3x the interval, then for the default situation you have a hold time of 75s, which is larger than the 60s timeout. I think you are thinking very specifically about your 2s interval, which I still insist is a very bad idea as it taxes the system way more than it should.

You also haven't explained why, in the default 25/60 setup, your ping packets take longer than 25s to be seen by the server, which I suppose is the whole reason you went on making these changes and reducing the interval?

christopher-abastar:
Hi Miguel,

> Maybe you don't realize that the client and the server do not rely on pings alone to detect disconnection. This is the method used when doing long-polling, but when the connection is over WebSocket, a disconnection is detected immediately, regardless how ping/ping packets.

Yes, you are right. We are using websockets only for app, as it isn't a browser base project.  So it did disconnects immediately.

> Do ping packets get delayed when you use the default 25/60 interval and timeout?
It really depends on the load. If there's nothing much going on the server, then ping packets should be ok. But if there's a sudden spike let's say at the 24s, a huge number of  connection, packet handling & broadcasting coming in, there is a high possibility that the PING packets will get delayed when processed by the server. 

> I think you are thinking very specifically about your 2s interval, which I still insist is a very bad idea as it taxes the system way more than it should.
> You also haven't explained why, in the default 25/60 setup, your ping packets take longer than 25s to be seen by the server, which I suppose is the whole reason you went on making these changes and reducing the interval?

I reduced the intervals purely because of testing (load test).  Reducing the interval, will simulate a multiple ping/connection handling and catch any load problems that I may encounter in the future.  So far,  I bumped into this issue when simulating a 1.5/msg per second on a t2.micro AWS (redis also installed inside) with 15-30 websocket clients, when it reach around 15 clients then it starts to disconnect some of them.  By my estimate, I would probably have 12.5x more if use  25/60 setup - So probably around 200+ clients before it starts disconnecting.  

Thanks!

miguelgrinberg:
@christopher-abastar I have changed the way the timeout works based on your proposal. But I went with something much simpler. Instead of using a 5/4 of the ping_interval, I just switched to the ping_timeout value. So now the time the server waits for a ping from the client with default settings is 60 seconds.

Let me know if that solves your problem.

christopher-abastar:
Thanks miguel!

christopher-abastar:
Hi miguel,

I saw some changes as well in latest revision in handling the sending of message packets, it was made to async, it looks ok so far on my test. 

However, I have another scenario happening as I played around with the ping_timeout (server sending a PONG).   I accidentally made it to a low_interval with 3x the ping_interval.

ping_interval = 3
ping_timeout = 9

When I tried to restart the server, some of the clients did an auto reconnect at the same time.  So, we are gonna be expecting the server to send the PONG at the **same moment in time**. I noticed some clients where disconnected frequently. Is it possible that the disconnects where happening because sending of PONGs where async=False?  I haven't sent any data in the server yet.

miguelgrinberg:
I'm still confused about the reasons you have to change these timeouts to such small numbers. The server sends the PONG immediately after receiving the PING packet sent by the client. Not sure if I understood you right, but you seem to imply there is a connection between ping_timeout and the time it takes for the server to send a PONG packet, but that is incorrect, the server does (and also did before these last changes) respond to a PING as fast as it can. If the server is slow to see the PING packet, that means that it is doing more than it can handle and the queue is backed up.

In that case the problem is the server load, and not how PING/PONG packets are handled. The possible solutions in that case is to simplify the work the handlers do, so that the queue can be serviced faster. That work can be moved to Celery worker processes, for example, which would take advantage of additional CPUs that you have. Another option that can be done on its own or in addition to Celery is to run multiple Socket.IO servers, so that the number of clients that are serviced by each process is lower.

christopher-abastar:
Hello Miguel,

I apologize, I wasn't clear on what was happening in the servers. After 0.9.2 I left everything to default values for the timer. However, my sentry server was popping some errors from time to time, and we only had around 8-10 connected clients at the same time.  So I had to tweak the timers to again to see if I can recreate the issue at higher frequency. So far, my investigation reveals that the server is receiving 2x disconnect event handler, not sure why.

I assure you that the workload is very minimal as it's just doing simple basic redis commands and no CPU/IO heavy at all. Though, i'm using the same redis for the pub/sub channel for the socket.io

Regards,

PR #25, opened 2016-08-03
Allow lower version of six

lukeyeager:
It seems that v1.5.2 is sufficient (because the tests pass). We could probably go lower, but [Ubuntu 14.04's packaged version](http://packages.ubuntu.com/trusty/python-six) seems like a nice standard.

Was there any particular reason to choose v1.9.0 originally?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     include_package_data=True,
     platforms='any',
     install_requires=[
-        'six>=1.9.0',
+        'six>=1.5.2',
     ],
     tests_require=[
         'mock',


miguelgrinberg:
Because that's the oldest version I tested everything with, the one that was current at the time. I'm really not a big fan of supporting older version of libraries. If I had to follow the same reasoning with every dependency I use my testing efforts would be much higher.

Following the reverse reasoning I can ask why do people need to install 2 or 3 year old libraries when it is as easy to install the latest versions.

lukeyeager:
> If I had to follow the same reasoning with every dependency I use my testing efforts would be much higher.

Actually I find it convenient to only support a single, stable version of a dependency rather than keeping up with all their updates. But I respect your choice too.

> Following the reverse reasoning I can ask why do people need to install 2 or 3 year old libraries when it is as easy to install the latest versions.

As stated, I like stable versions of things. Also, not all of my dependencies are Python packages - some are binaries and/or shared libraries. Deb packages work well for all of my dependencies and enable a super easy installation mechanism for users. So, when there's already a deb package out for v1.5.2, I prefer not to have to make a new package for v1.9.0 if I don't have to.

But I recognize that you and most of your users don't care about this. I'll figure out how to work around it on my own. Thanks for your consideration.

miguelgrinberg:
> As stated, I like stable versions of things

Stable according to who? All the dependencies I use are stable, I don't work with alpha/beta/unstable releases at all. Why is Ubuntu better at deciding what's stable than the own package owners?

If you install the Ubuntu blessed `python-pip` package, then you can selectively install a combination of "stable" releases out of Ubuntu or Pypi, as necessary to satisfy the dependency requirements of the applications that you use.

lukeyeager:
This is probably not a very productive argument. [I've ](https://github.com/NVIDIA/DIGITS/commit/db35cd84c607db0d1b0679327acb3ff5b07618ad) [just](https://github.com/NVIDIA/DIGITS/pull/415) [had](https://github.com/NVIDIA/DIGITS/pull/640) [some](https://github.com/NVIDIA/DIGITS/pull/648) [bad](https://github.com/NVIDIA/DIGITS/pull/824) [luck](https://github.com/NVIDIA/DIGITS/issues/929) with the pip ecosystem.

miguelgrinberg:
I routinely experience the same kind of problems. As new release come out there's always problems that need to be addressed. I'm not sure why you attribute that to the pip ecosystem though. I doubt the arbitrary versions selected by Ubuntu to include in their distros are any more reliable than any versions you can get from pip.

Basically you are advocating for "pinning" all packages to known good versions, and you test your software with Ubuntu's chosen versions and declare those your known good versions. That is a totally understandable working model, but the problem is that at the same time you want the greatest and latest Flask-SocketIO. If you want to have everything pinned to 2-3 years ago then install a version of Flask-SocketIO from that time, something like the 0.6 release, which was based on gevent-socketio, and you'll have a lot less problems with dependencies.

lukeyeager:
In general deb packages work great for me. In general I'm **not** interested in the "greatest and latest" for dependencies. The only reason I'm bringing any of this up is because I need to upgrade Flask-SocketIO in order to support Ubuntu 16.04 because of the gevent update and now I'm running into a naming conflict, as laid out here: https://github.com/miguelgrinberg/python-socketio/issues/23#issuecomment-237073926. That's my real problem. This `six` version issue for `python-engineio` is very minor - I've already [overridden it](https://github.com/lukeyeager/python-engineio/blob/debian/0.9.2/debian/patches/six-version.patch) and [built a package](https://launchpad.net/~luke-yeager/+archive/ubuntu/digits).

I understand that you prefer staying up-to-date and pip works great for that. It doesn't suit my use case and I'm not switching to it. I also understand that you're not interested in supporting debian packaging efforts, which is totally your prerogative. I'll take the hint and stop bugging you with these sorts of compatibility requests.

As always, thanks for the great software you produce!

miguelgrinberg:
> I also understand that you're not interested in supporting debian packaging efforts, which is totally your prerogative
I don't mind supporting packagers of distros, and I actually do when it does not add to my efforts. But supporting today's version of my software against couple year old dependencies I never tested it against is not something that I am willing to support.

PR #26, opened 2016-08-18
Uwsgi gevent support

efficiosoft:
I added websocket support for gevent with uWSGI using uWSGI's native websocket api. There is currently no detailed documentation for this update, but code comments should suffice for you to decide whether to merge this into master or not. Feedback is heavily appreciated. Best regards.

### DIFF
--- a/engineio/async_gevent.py
+++ b/engineio/async_gevent.py
@@ -125,7 +125,6 @@ async = {
     'queue_class': 'JoinableQueue',
     'websocket': sys.modules[__name__] if _websocket_available else None,
     'websocket_class': {'gevent': 'GeventWebSocket', 'uwsgi': 'uWSGIWebSocket',
-                        False: None}
-                       [_websocket_available],
+                        False: None}[_websocket_available],
     'sleep': gevent.sleep
 }
--- a/engineio/async_gevent.py
+++ b/engineio/async_gevent.py
@@ -4,7 +4,10 @@ import sys
 import gevent
 try:
     import uwsgi
-    _websocket_available = "uwsgi"
+    if hasattr(uwsgi, 'websocket_handshake'):
+        _websocket_available = "uwsgi"
+    else:
+        raise ImportError('uWSGI not running with websocket support')
 except ImportError:
     try:
         import geventwebsocket  # noqa
@@ -122,7 +125,7 @@ async = {
     'queue_class': 'JoinableQueue',
     'websocket': sys.modules[__name__] if _websocket_available else None,
     'websocket_class': {'gevent': 'GeventWebSocket', 'uwsgi': 'uWSGIWebSocket',
-                        False: None} \
+                        False: None}
                        [_websocket_available],
     'sleep': gevent.sleep
 }
--- a/engineio/async_gevent.py
+++ b/engineio/async_gevent.py
@@ -97,12 +97,13 @@ class uWSGIWebSocket(object):  # pragma: no cover
                     gevent.select.select([fd], [], [])[0]
             self._select_greenlet = gevent.spawn(
                 select_greenlet_runner,
-                uwsgi.connection_fd(**self._uwsgi_api_kwargs),
+                uwsgi.connection_fd(),
                 self._event)
 
         return self.app(self)
 
     def close(self):
+        uwsgi.disconnect()
         if self._event is not None:
             self._select_greenlet.kill()
             self._event.set()
--- a/engineio/async_gevent.py
+++ b/engineio/async_gevent.py
@@ -74,13 +74,10 @@ class uWSGIWebSocket(object):  # pragma: no cover
 
         uwsgi.websocket_handshake()
 
-        self._event = None
-        self._send_queue = None
-        self._select_greenlet = None
-        self._uwsgi_api_kwargs = {}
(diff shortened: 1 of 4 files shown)


efficiosoft:
Sorry for these travis-ci build failures. I'm currently adjusting the tests accordingly.

efficiosoft:
Ok, tests fixed now and all travis-ci checks pass. Sorry again for that delay. Would love to hear your feedback.

miguelgrinberg:
This is awesome, thanks a lot. I will test in the next few days and let you know if I find any problems.

engineio/async_gevent.py
```diff
@@ -52,12 +56,71 @@ def wait(self):
...
+            return None
+        return msg.decode()
+
+    def send_and_wait(self, poll_func):
+        """Calls poll_func(block=False) regularly and sends the messages
+        it returns. Apart from that, it observes the websocket for incoming
+        messages. If one arrives, it returns that. After the returned message
+        was processed, this function can be called again to continue its work.
+        A return of None means that the connection was closed. Afterwards."""
+        while True:
```
miguelgrinberg: I'm going to test this, but from the looks of it, this is going to consume CPU, whether there is data traffic or not.  I prefer the approach I'm using, which lets the I/O functions block in another thread until there is something to do.

efficiosoft:
@miguelgrinberg Yes, this is preferable. But what I didn't mention: uWSGI websocket api can only be used from the main greenlet. In another project I used gevent.select on both the connection's and a redis's file descriptor that was used to sync events between servers. But since we have a JoinableQueue here instead of redis, I'm not sure whether select can be used. Maybe one can get the file descriptor of the JoinableQueue's pipe and call select on that?

efficiosoft:
BTW: JoinableQueue._reader works with select.select(), but it's a private attribute and thus not part of the official multiprocessing api. However, I could change the implementation to work with it if you prefer. But I'm not sure about how this would perform on Windows because .fileno() doesn't return a file descriptor there.

efficiosoft:
I committed a new approach that lacks the socket polling at the cost of an extra greenlet and an extra queue per websocket. Should be reasonable I think.

miguelgrinberg:
Thanks, this looks much better. :)

Do you have a link to where uWSGI documents the limitations with calling the websocket functions from a greenlet that is not the main one? I could not find any mention of that in the docs.

efficiosoft:
I couldn't find it in the docs either, but their github issues are full of "[BUG] current_wsgi_req NOT FOUND !!!".
This is valid for the whole uwsgi python api, including functions like uwsgi.connection_fd(). Unfortunately, the documentation for the uwsgi python module is less than rare, but I could reproduce the current_wsgi_req not found bug mentioned above.

Some resources mentioning it:
https://github.com/unbit/uwsgi/issues/980
https://github.com/unbit/uwsgi/blob/master/plugins/gevent/gevent.c

But however their documentation is not the best, uwsgi is one of the most performant and stable wsgi container servers.

efficiosoft:
This version now dynamically chooses the most performant approach depending on the uWSGI version available. I'll probably leave it this way. Tested it with stable (2.0.13.1) and development (2.1-dev-907dbd5). Both work.

efficiosoft:
@miguelgrinberg I'm still unsure about the encoding to pass to msg.decode() on messages received over websocket. uWSGI apparently provides them as bytes, but engineio expects them to be of type str. AFAIK websockets use UTF-8 when operated in non-binary mode So I guess I should specify 'utf-8' explicitly when decoding received messages, right?
On the other hand, engineio passes str objects into send(). uWSGI's send function accepts them, but probably they should be encoded with UTF-8 as well before passing them to uWSGI for transmission?

efficiosoft:
Thoughts about encoding and decoding are cleared now. Binary support is included and working. Please let me know if there is something else I could implement or have to test.

miguelgrinberg:
@efficiosoft what's the uwsgi command line that you use to start the server?

efficiosoft:
This is my current uwsgi.yaml:

```
uwsgi:
    gevent: 1000
    processes: 1
    listen: 128
    http-socket: 127.0.0.1:8080
    static-map: /static=./static
    static-index: index.html
    module: myapp.wsgi
    logformat: %(status) - %(method) %(uri) - %(msecs) ms
    harakiri-verbose: 1
    no-orphans: 1
    vacuum: 1
    master: 1
    enable-threads: 1
    touch-reload: ./uwsgi.reload
```

miguelgrinberg:
Okay, got it to work with the following command line:

    uwsgi --http :5000 --wsgi-file latency.py --callable app --gevent 1000 --http-websockets --master

Thanks.

miguelgrinberg:
Seems to work well, so I just merged these changes, after I made a bit of restructuring. The main difference is that I moved the gevent+uwsgi support to a separate async mode, so now have to specify `async_mode='gevent_uwsgi'`. With this change, in most cases you don't even need to explicitly set the async mode, as it will be selected automatically if the correct dependencies are installed.

efficiosoft:
Thanks a lot for merging! Splitting it into two modules seems reasonable. I updated the doc string.for `Server` class because it didn't reflect the new `async_mode` yet. You might want to merge this as well.

PR #27, opened 2016-08-21
Optionally pass environ to message event handler

efficiosoft:
I was looking for a method to always have the current `environ` at hand in an event handler. For that to work, I added the `pass_environ` keyword argument to `engineio.Server` initialization with a default of `False`. If it is set to `True`, the current `environ` is passed to the event handler for `'message'` as second argument, after `sid`.

By _current_ `environ` I mean the one belonging to a particular event triggering.

This behaviour ensures backwards compatibility and is not noticeable for those who don't enable it explicitly.

I also wrote an addition to the `socketio` library to let it pass through the `environ` to it's registered event handlers. I'll submit another pull request for that part.

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -50,6 +50,9 @@ class Server(object):
     :param async_handlers: If set to ``True``, run message event handlers in
                            non-blocking threads. To run handlers synchronously,
                            set to ``False``. The default is ``True``.
+    :param pass_environ: If set to ``True``, the current ``environ`` is passed
+                         to the ``"message"`` event handler as second argument
+                         (right after ``sid``). (default is ``False``)
     :param kwargs: Reserved for future extensions, any additional parameters
                    given as keyword arguments will be silently ignored.
     """
@@ -61,7 +64,7 @@ class Server(object):
                  http_compression=True, compression_threshold=1024,
                  cookie='io', cors_allowed_origins=None,
                  cors_credentials=True, logger=False, json=None,
-                 async_handlers=True, **kwargs):
+                 async_handlers=True, pass_environ=False, **kwargs):
         self.ping_timeout = ping_timeout
         self.ping_interval = ping_interval
         self.max_http_buffer_size = max_http_buffer_size
@@ -72,6 +75,7 @@ class Server(object):
         self.cors_allowed_origins = cors_allowed_origins
         self.cors_credentials = cors_credentials
         self.async_handlers = async_handlers
+        self.pass_environ = pass_environ
         self.sockets = {}
         self.handlers = {}
         if json is not None:
@@ -230,8 +234,6 @@ class Server(object):
                         self.logger.warning('Invalid session %s', sid)
                         r = self._bad_request()
                     else:
-                        self._trigger_event('connect_update', sid, environ,
-                                            async=False)
                         socket = self._get_socket(sid)
                         try:
                             packets = socket.handle_get_request(
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -230,6 +230,8 @@ class Server(object):
                         self.logger.warning('Invalid session %s', sid)
                         r = self._bad_request()
                     else:
+                        self._trigger_event('connect_update', sid, environ,
+                                            async=False)
                         socket = self._get_socket(sid)
                         try:
                             packets = socket.handle_get_request(
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -35,7 +35,7 @@ class Socket(object):
             pass
         return packets
 
-    def receive(self, pkt):
+    def receive(self, pkt, environ=None):
         """Receive packet from the client."""
         self.server.logger.info('%s: Received packet %s data %s',
                                 self.sid, packet.packet_names[pkt.packet_type],
@@ -45,8 +45,13 @@ class Socket(object):
             self.last_ping = time.time()
             self.send(packet.Packet(packet.PONG, pkt.data))
         elif pkt.packet_type == packet.MESSAGE:
-            self.server._trigger_event('message', self.sid, pkt.data,
-                                       async=self.server.async_handlers)
+            if self.server.pass_environ:
+                self.server._trigger_event('message', self.sid, environ,
+                                           pkt.data,
+                                           async=self.server.async_handlers)
+            else:
+                self.server._trigger_event('message', self.sid, pkt.data,
+                                           async=self.server.async_handlers)
         elif pkt.packet_type == packet.UPGRADE:
             self.send(packet.Packet(packet.NOOP))
         else:
@@ -94,7 +99,7 @@ class Socket(object):
             body = environ['wsgi.input'].read(length)
             p = payload.Payload(encoded_payload=body)
             for pkt in p.packets:
-                self.receive(pkt)
+                self.receive(pkt, environ=environ)
 
     def close(self, wait=True, abort=False):
         """Close the socket connection."""
@@ -182,7 +187,7 @@ class Socket(object):
                 p = p.encode('utf-8')
             pkt = packet.Packet(encoded_packet=p)
             try:
-                self.receive(pkt)
+                self.receive(pkt, environ=ws.environ)
             except ValueError:
                 pass
 
--- a/tests/test_socket.py
+++ b/tests/test_socket.py
@@ -21,6 +21,7 @@ class TestSocket(unittest.TestCase):
         mock_server.ping_timeout = 0.2
         mock_server.ping_interval = 0.2
         mock_server.async_handlers = True
+        mock_server.pass_environ = False
 
         try:
             import queue


miguelgrinberg:
Doesn't `request.environ` give you what you want?

efficiosoft:
What is request and where does it come from? I'm talking about raw engineio, not django.

Am 22.08.2016 um 03:47 schrieb Miguel Grinberg <notifications@github.com<mailto:notifications@github.com>>:


Doesn't request.environ give you what you want?

-

miguelgrinberg:
I get that you are talking about raw engineio, but there is no great reason to use engineio alone. And even if you do that, there is already a way for an application to access `environ`, as it gets it in the connect event handler, where it can store it for later use.

The idea of passing the environments for every request that the client sends does not really go well with the way the Socket.IO protocol works. At any given time, a client using long-polling could have up to two outstanding requests, a GET request to read, and a POST request to write. In this case there could be two different `environ` dictionaries, so which one do you send to handlers? And then when the connection migrates to websocket, there are no more requests, so what do you use then, given that the only environment that you have is now stale?

efficiosoft:
The thing here is that every event handler might want to have access to an environ that stays the same through the whole process of handling that event. In case of websockets, the environ will start to be the same on every event as soon as the connection upgrade is finished. But it won't be the environ stored in ``socketio.Server.environ[sid]``, because that is the one that was recent at connection time and is now outdated for sure. My approach provides each event handler triggered with the ``environ`` that caused it to be triggered. BTW: WebSocket environs are not stale, they are just valid for a longer time, aren't they?

For long polling you are right. Currently, it passes the POST request's ``environ`` that triggered the event because the GET request might not yet be available at that time. It would be better to wait for the GET request to arrive in that case before starting the event handler, right?

The reason why I wrote that extension is the following:

I wanted to use Django's sessions that rely on cookies being read and injected in the request/response cycle. If I now for example authenticate a client in an ``socketio`` event handler, its session gets altered and also the cookie's value that is going to be transmitted to the client together with the HTTP response.
Until now, I only had the ``environ`` available that was valid at Engine.IO connection time. But the response belonging to that ``environ`` had already been sent and cookie injection is not possible anymore.
With my extension enabled, I now have the ``environ`` that belongs to the current request (however it's not yet the one of the reading request when using long-polling).

efficiosoft:
Ok, I see that it isn't possible to map GET requests to POST requests. So ``pass_environ=True`` provides valid environs in the following situations:

1. WebSockets are used. The ``environ`` is valid throughout the whole lifetime of the WebSocket connection.
- Long polling is used. The environ provides information to authenticate the user or read cookies he transmitted together with his request, for instance. Of course, other decisions could be made like blacklisting specific events by ``environ['REMOTE_ADDR']`` correctly even if the users IP has changed since he initially connected.

miguelgrinberg:
> WebSockets are used. The environ is valid throughout the whole lifetime of the WebSocket connection and can be used .unrestrainedly for everything (reading and sending cookies etc.).

Can you explain this in more detail? How do you send a cookie through a WebSocket connection?

efficiosoft:
You are right, sending cookies through web socket connections is not possible. I must have been sleeping when I was writing this.

However, there is a legit use case for an up-to-date ``environ``: authorization, let it be based on cookies, ``REMOTE_ADDR`` or something else. You can not send cookies through both transports, but you can read metadata that is as recent as possible.

miguelgrinberg:
You do not need to authenticate every event. Authenticating and authorizing on connect is sufficient, given that the connection with the client is stateful.

efficiosoft:
But if the client authenticates itself at some point within the socketio connection, for example by triggering a login event, currently the only way to remember that he has authenticated is to store this information together with his sid somewhere. And if you are running multiple workers, you also need to synchronize them.

By having a recent environment available, you can let your framework of choice do session handling for you. And this usually works for multiple workers out of the box. Django for instance can store whatever you want in a session and make that data available in subsequent requests no matter what worker they run in.

miguelgrinberg:
Sorry, but I still don't see what your plan is. As I said above, you can't assume that the request environment is recent. For a websocket connection you will not be able to get environment updates. For example, if cookies get updated through HTTP, you will not see those updates. There is no easy way to store a user session, at least I'm not aware of any that work via the request environment.

efficiosoft:
You can't send cookies via web socket, yes. But if you use web socket, you will never change the worker while the connection is running and thus don't need to share a session between requests.

There is no way to store a user session in environment I'm aware of. But your web framework of choice can take a cookie from the environment to perform a database query, for instance, in order to fetch a session that can be altered and stored again.

efficiosoft:
How do you actually share a users state between workers when using polling? With an environment to take a cookie from it would be easy.

efficiosoft:
I've got another idea. How about providing a way to to overload sid generation based on environ? I could then take the same sid as The one chosen by django for its session. Then, the sid would be enough information to fetch the session.

miguelgrinberg:
> How do you actually share a users state between workers when using polling? With an environment to take a cookie from it would be easy.

You can solve the problem only when using polling. At any given time you will have a bunch of clients using long-polling, and another bunch using websocket.

You should take a look at my Flask-SocketIO project. That's the high-level Socket.IO server for Flask that I wrote based on python-engineio and python-socketio.

Not sure exactly what you need to share in your application. For the internal data that the Socket.IO server stores for each user I do not need to share it across workers. The load balancer is configured for sticky sessions, so a given user is always managed by the same worker. In the event that a worker needs to be taken offline, then the users that were managed by the worker need to reconnect to be assigned a new worker.

efficiosoft:
When sessions stick to workers, this is a feasible approach. However sessions are not persistent over server restarts when only stored in memory. I personally like the database backed way that also works accross workers. Why can't we give every event handler an ``environ`` if it wants to have it? I see no other way of sharing something user-specific between workers unless you use sessions based on the ``sid``. However, this would kind of mangle django's request processing cycle where it loads the session based on cookie and fetches the current user from database.

efficiosoft:
Maybe I just overwrite the django session key with the ``sid`` taken from query string early in the request processing cycle. This would also work across workers and ensure everybody uses the same session.

miguelgrinberg:
Not sure you realize this, but the idea of moving requests across workers is not going to work. It is required to use sticky sessions, as stated in the documentation.

efficiosoft:
Oh! No, I didn't realize this and wonder why I didn't see it at my deep dive into your Engine.IO and Socket.IO libraries during the last days.

I'll see how to implement that load balancing correctly, since using the hashed IP doesn't seem useful when looking at how e.g. mobile broadband providers change your IP every now and then, sometimes even on a per connection basis. May be I'll let the load balancer learn the different sids generated and map them to workers. But this is another unrelated issue.

So I think I'll use the ``socketio.Server().environ`` dictionary to access the environments stored at connection time for authentication. And to update cookies from event handlers, I'm going to use a ``setCookie`` event triggered by the server. The next time the client connects, it's sending the updated cookies and thus will be authenticated automatically... Sounds doable.

So do you finally suggest trashing the ``pass_environ`` parameter now?

miguelgrinberg:
Link to deployment of multiple worker processes: https://python-socketio.readthedocs.io/en/latest/#multi-process-deployments.

I really see no benefit of having mid-connection WSGI environments, so far I haven't seen any compelling reason to implement this option. The different ways in which long-polling and websocket transports are implemented make the use of environments at times other than connection unreliable.

> And to update cookies from event handlers

I believe we discussed this above. You certainly understand your application much better than I do, but updating cookies from event handlers is not a good idea. You are not guaranteed that an event handler will trigger an HTTP response back to the client. I think you need to stop thinking about events in terms of HTTP requests. They are not the same, the rules are different.

efficiosoft:
@miguelgrinberg Yes, I got it. As I wrote, I'm going to set cookies manually at client-side, triggered by a server-emitted event, not via HTTP responses.

Thank you a lot for this detailed discussion. I really liked it. And with the sticky-session thing in mind, I think the two pull requests can be closed. But since I'm going to work with your library the next few months, there may come others. Again,

miguelgrinberg:
@efficiosoft you should take a look at my Flask-SocketIO extension. I have thought about all these problems and have made sure there are solutions for them for Flask. In particular I document how to work with Flask-Login, the most used authentication extension, which also uses cookies to store the logged in state of users.

efficiosoft:
The problem in my particular case is that I'm kind of forced to use django because there is a relatively large ORM already written with it. I see the benefits of a lightweight framework like flask. Maybe I'm really starting to port it all to sqlalchemy and use flask, but I'm not sure, yet.

miguelgrinberg:
@efficiosoft no sorry, didn't mean to suggest you should drop django. Just that you can take ideas from the solutions I built for Flask.

efficiosoft:
Yes, I'm definitely having a closer look at it. Thanks.

I scetched the pattern I'm going to use now and it actually works like I expected with django's sessions. Sure, I use the stale request object built at connection time but it has the session attached to it and can be used for authorization and logging in and out. Great.

PR #31, opened 2017-01-02
fix the bytearray has not .encode() error

wwqgtxx:


miguelgrinberg:
@wwqgtxx what payload did you send from the client that ended up being a bytearray on the server side?

wwqgtxx:
if the websocket opcode is 0x2

wwqgtxx:
As we know,the socket.io protocol alow the binary data send by two way.One is send a base64encode data,the other one is send a raw bytes data.If want to send a websocket raw bytes data, It need to set websocket opcode to 0x2,than the python-engineio had receive the bytearray data from client.

miguelgrinberg:
Sorry, but it seems you are confusing websocket with Socket.IO. My question was which Socket.IO client sends packets that on the server appear as byte arrays. All the tests I've done with the official Socket.IO clients send bytes, not byte arrays. 

If you are using a plan websocket client and expecting Socket.IO on the server will understand that then that's the problem, these two protocols are not compatible, you need to use a Socket.IO client.

wwqgtxx:
I am trying to make python's socketIO-client support binary data send,and add this pr to socketIO-client
https://github.com/invisibleroads/socketIO-client/pull/144
I had test that if I send a base64 encode binay data, everythings will be fine.But it well waste many space because of base64.So I try to send the same package like python-socketio and it cause the python-engineio recevied the bytearray data.

miguelgrinberg:
Ah, okay, that makes sense. So this is the python Socket.IO client.

On the server, are you using eventlet, gevent or uwsgi?

wwqgtxx:
I use gevent.

wwqgtxx:
I found this problem cause by `geventsebsocket`:
```python
    def read_message(self):
        """
        Return the next text or binary message from the socket.

        This is an internal method as calling this will not cleanup correctly
        if an exception is called. Use `receive` instead.
        """
        opcode = None
        message = bytearray()

        while True:
            header, payload = self.read_frame()
            f_opcode = header.opcode

            if f_opcode in (self.OPCODE_TEXT, self.OPCODE_BINARY):
                # a new frame
                if opcode:
                    raise ProtocolError("The opcode in non-fin frame is "
                                        "expected to be zero, got "
                                        "{0!r}".format(f_opcode))

                # Start reading a new message, reset the validator
                self.utf8validator.reset()
                self.utf8validate_last = (True, True, 0, 0)

                opcode = f_opcode

            elif f_opcode == self.OPCODE_CONTINUATION:
                if not opcode:
                    raise ProtocolError("Unexpected frame with opcode=0")

            elif f_opcode == self.OPCODE_PING:
                self.handle_ping(header, payload)
                continue

            elif f_opcode == self.OPCODE_PONG:
                self.handle_pong(header, payload)
                continue

            elif f_opcode == self.OPCODE_CLOSE:
                self.handle_close(header, payload)
                return

            else:
                raise ProtocolError("Unexpected opcode={0!r}".format(f_opcode))

            if opcode == self.OPCODE_TEXT:
                self.validate_utf8(payload)

            message += payload

            if header.fin:
                break

        if opcode == self.OPCODE_TEXT:
            self.validate_utf8(message)
            return str(message, "utf-8")
        else:
            return message
```
because it recevie a `OPCODE_BINARY`  so it will return a `bytearray` without decode to str

wwqgtxx:
Above `geventwebsocket` is from package `karellen-geventws` because I use python3.5.
But I see the origin `geventwebsocket` do the same things in the code
```python
    def read_message(self):
        """
        Return the next text or binary message from the socket.

        This is an internal method as calling this will not cleanup correctly
        if an exception is called. Use `receive` instead.
        """
        opcode = None
        message = ""

        while True:
            header, payload = self.read_frame()
            f_opcode = header.opcode

            if f_opcode in (self.OPCODE_TEXT, self.OPCODE_BINARY):
                # a new frame
                if opcode:
                    raise ProtocolError("The opcode in non-fin frame is "
                                        "expected to be zero, got "
                                        "{0!r}".format(f_opcode))

                # Start reading a new message, reset the validator
                self.utf8validator.reset()
                self.utf8validate_last = (True, True, 0, 0)

                opcode = f_opcode

            elif f_opcode == self.OPCODE_CONTINUATION:
                if not opcode:
                    raise ProtocolError("Unexpected frame with opcode=0")

            elif f_opcode == self.OPCODE_PING:
                self.handle_ping(header, payload)
                continue

            elif f_opcode == self.OPCODE_PONG:
                self.handle_pong(header, payload)
                continue

            elif f_opcode == self.OPCODE_CLOSE:
                self.handle_close(header, payload)
                return

            else:
                raise ProtocolError("Unexpected opcode={0!r}".format(f_opcode))

            if opcode == self.OPCODE_TEXT:
                self.validate_utf8(payload)

            message += payload

            if header.fin:
                break

        if opcode == self.OPCODE_TEXT:
            self.validate_utf8(message)
            return message
        else:
            return bytearray(message)
```

miguelgrinberg:
Okay, that explains it. Thanks for the detailed analysis. I did not find this problem myself because I rarely use gevent-websocket for my testing, since it has never been ported to Python 3. I should give the 3.5 fork a try, I guess.

miguelgrinberg:
@wwqgtxx Thanks, the PR is now merged. I have added a couple other places where binary types were being checked, plus some unit tests.

wwqgtxx:
Only a suggest that you can try to use [ws4py](https://pypi.python.org/pypi/ws4py) package to support the gevent on both python2 and python3.
I had read this package's doc and think it can provide all the operater that python-engineio will used.

miguelgrinberg:
Thanks, I will take a look at ws4py and see if I can integrate it.

Issue #30, opened 2016-12-13
EngineIO fails


fokinpv:
Hello, 

I'm using Flask-SocketIO package and from time to time it's failed with the such traceback [SocketIO Traceback](https://gist.github.com/fokinpv/b4b9a9b59eba93328e28d5f95c01ca97). I couldn't find out specific conditions, it randomly happens on the same data passing thru websockets.

This appears as with eventlet as with using of gevent.

Here is [pip freeze](https://gist.github.com/fokinpv/82f44ea95a6b8606225f8fef3015b5a9).

What additional information may help? What can it be?

miguelgrinberg:
What client are you using? The error suggests the client is using a different version of the engine.io protocol with different packet types.

fokinpv:
I'm using socket.io-client: 1.4.5. Is it ok? I've just seen that in flask-socketio examples is 1.3.6

miguelgrinberg:
Version 1.4.5 should be fine, and in fact, even newer versions are fine too, I see no changes in the packet types.

You are not writing to the socket manually, correct? You are sending all your messages via the `emit()` function in the client? It's odd that the server is receiving an invalid packet type.

Can I ask you to do a small experiment for me? Open file `/home/user/Envs/venv/local/lib/python2.7/site-packages/engineio/socket.py` in any text editor, and change line 42 from this:

    self.sid, packet.packet_names[pkt.packet_type],

to this:

    self.sid, str(pkt.packet_type),

Then in your application, in the place where you create your `SocketIO` object, please add the following argument: `engineio_logger=True`.

That will take care of the error you are receiving, and also enable more verbose logging. Run the application again, and if you still get errors (I think you will, but on a different line now) then send me the logs all the way to, and including, the error.

After you capture a session with an error in it, you can revert that line 42 to the original contents.

fokinpv:
Could not reproduce it.

PR #37, opened 2017-02-13
Fix crash on invalid packet type. Add test

voronin-opensoft:


engineio/socket.py
```diff
@@ -41,8 +41,10 @@ def poll(self):
 
     def receive(self, pkt):
         """Receive packet from the client."""
+        packet_name = packet.packet_names[pkt.packet_type] \
+            if pkt.packet_type in packet.packet_names else ''
```
AlexElvers: This will always return an empty string. `pkt.packet_type` is an int, `packet.packet_names` is a list of strings.

dimka665: Yes. Thanks

engineio/socket.py
```diff
@@ -41,8 +41,10 @@ def poll(self):
 
     def receive(self, pkt):
         """Receive packet from the client."""
+        packet_name = packet.packet_names[pkt.packet_type] \
+            if pkt.packet_type <= len(packet.packet_names) else 'UNKNOWN'
```
miguelgrinberg: Shouldn't this be "less" instead of "less than"?

miguelgrinberg:
Out of curiosity, what client are you using that feeds invalid packet types to the server?

voronin-opensoft:
Hello.
It fixes crash on invalid packet type. It solves problem from #30 issue . We uses `socket.io-client: 1.4.5` and gets WebSocket packet consist of one byte with random value: https://github.com/miguelgrinberg/python-engineio/blob/master/engineio/socket.py#L190

miguelgrinberg:
Thanks!

PR #51, opened 2017-06-07
fix memory leak

net21:
the items in queue doesn't release when socket close.

### DIFF
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -120,6 +120,11 @@ class Socket(object):
             self.closed = True
             if wait:
                 self.queue.join()
+            else:
+                # release the items in queue
+                while not self.queue.empty():
+                    self.queue.get()
+                    self.queue.task_done()
             if reraise_exc:
                 six.reraise(*reraise_exc)
 


miguelgrinberg:
I don't understand the purpose of this fix. I don't believe this is a memory leak. The socket is going to close and will be deleted, and with it, the queue will be deleted as well. Or am I missing something?

net21:
i use tracemalloc to trace the memory used, i emit many object to socket every second, the memory used don't released when client lose the connection, and i checked the close method is called, but the objects in queue don't released , so i have to do it .

net21:
Any of the "del self.sockets[sid]" sentence in engineio\server.py wasn't be called.

miguelgrinberg:
> Any of the "del self.sockets[sid]" sentence in engineio\server.py wasn't be called.

Wouldn't that also cause the socket object and the queue to be leaked? You said just the items in the queue were leaked, or did I misunderstand?

net21:
As you said, the socket object and the queue all will be leaked. It's the new found.

miguelgrinberg:
So then this PR is not a complete fix? There's still leaked objects?

net21:
The server object don't have any chance to del the socket if client don't send the  disconnect  event in polling mode.

miguelgrinberg:
That is incorrect. Any time a message is emitted to a timed out socket, the server will delete it. Is that the situation you are trying to address? I think a better solution would be to be more efficient in detecting these sockets that are abandoned, instead of waiting until they need to be used.

net21:
I has add some log code, the close method in socket object has been called, but the socket object seems never be deleted in server object.

miguelgrinberg:
Do you have a stack trace for the call to `close()`?

net21:
File "C:\Python34\lib\site-packages\flask_socketio\__init__.py", line 346, in emit
    skip_sid=skip_sid, callback=callback, **kwargs)
  File "C:\Python34\lib\site-packages\socketio\server.py", line 221, in emit
    **kwargs)
  File "C:\Python34\lib\site-packages\socketio\base_manager.py", line 135, in emit
    self.server._emit_internal(sid, event, data, namespace, id)
  File "C:\Python34\lib\site-packages\socketio\server.py", line 396, in _emit_internal
    binary=binary))
  File "C:\Python34\lib\site-packages\socketio\server.py", line 407, in _send_packet
    self.eio.send(sid, encoded_packet, binary=False)
  File "C:\Python34\lib\site-packages\engineio\server.py", line 193, in send
    socket.send(packet.Packet(packet.MESSAGE, data=data, binary=binary))
  File "C:\Python34\lib\site-packages\engineio\socket.py", line 72, in send
    self.close(wait=False, abort=True)
  File "C:\Python34\lib\site-packages\engineio\socket.py", line 111, in close
    traceback.print_stack()

miguelgrinberg:
This was fixed in a different way now.

Issue #62, opened 2018-02-12
engineio incomaptible with aiohttp-3


maksbotan:
Hi! While debugging strange behaviour of Socket.IO with aiohttp server I've come across an issue in engineio.

You are calling `aiohttp.web.WebSocketResponse.send_str()` in `async_aiohttp.py`, but since aiohttp-3 this method is actually a coroutine, so you have to await it:

https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.WebSocketResponse.send_str

Can you please fix this? I can file a simple PR adding `await` and `aiohttp>3` requirement, but I guess you'd want to support both versions.

Right now I had to downgrade `aiohttp` on my server to workaround this.

stephan48:
i am also affected by this bug. would it be possible to detect what type send_str is(function or coroutine) and await it conditionally?

miguelgrinberg:
Yes, it is possible, that is actually how I'm going to fix it. Just did not get around to do it yet.

stephan48:
Hi @miguelgrinberg,

Since my app requires aiohttp>=3,  I went ahead and proposed & tested a fix. Kind of wanted to test that little part of my work project before I go to bed.

Sadly i can't test it on aiohttp 2 atm due to time constraints. But i expect it to work - unless i forgot some Python certicularity. Please review and comment on my PR and maybe merge it, if it fits your liking :)

Thank you for you work!

--Stephan

larry-dev:
@miguelgrinberg  When will the problem be repaired?  Thank you very much!

miguelgrinberg:
Could any of you give the master branch a try? I think all the problems with aiohttp3 should be resolved now. If everything seems to be okay I'll cut a new release. Thanks!

Kruptein:
I did a quick local update from the engineio master branch and the error is no longer being logged and websockets are being used instead of the polling mechanism.

miguelgrinberg:
Fix released in v2.0.3.

PR #63, opened 2018-02-14
aiohttp>=3 expects the WebSocketResponse.send_str and WebSocketRespon…

stephan48:
…se.send_bytes functions to be awaited(coroutines) - do this in a backwards compatible way

A proposed fix for Bug #62.

### DIFF
--- a/engineio/async_aiohttp.py
+++ b/engineio/async_aiohttp.py
@@ -101,7 +101,7 @@ class WebSocket(object):  # pragma: no cover
 
     async def send(self, message):
         async def call_async_conditionally(func, arguments):
-            if inspect.iscoroutinefunction(object):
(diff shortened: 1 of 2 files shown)


stephan48:
I think theres another regression with aiohttp>=3 - but according to travis it looks like it works with aiohttp<=2

stephan48:
Travis is being weird. it does not really want to build my second commit - even on a rebuild it seems to spit out test errors - which are not fixed in my first commit. Weird thing - the tests pass for me atleast :)

The mentioned regression happens in https://github.com/aio-libs/aiohttp/commit/e0bb501c15ff88c579fe7c97490c4e8e6af875fc

miguelgrinberg:
This has been fixed. Closing.

Issue #56, opened 2017-06-28
Race condition causes file descriptors leak 


milenyordanov:
I have a simple 1:1 chat application.
It is built with these components:

Flask==0.12.1
Flask-SocketIO==2.9.0
python-engineio==1.6.1
python-socketio==1.7.5

It uses async_mode='eventlet'.

The chat clients connect to the server and exchange messages. If a chat client is disconnected then the server sends a message to the other client in the room.

The file descriptors leak happens in the following case.
- Both chat clients are joined in a room.
- Both chat clients are created in a single process. This process is a test application.
- The test application process is killed.
- The server receives events 'disconnect' for both clients.

The 'disconnect' event handler code is something like this:

        @socketio.on('disconnect')
        def client_disconnect():
                emit('sig-user-leaved', {'client_id': client_id}, room=room, include_self=False)

I traced the problem in the following code: socket.py end of the method _websocket_handler.

        self.queue.put(None)  # unlock the writer task so that it can exit
        writer_task.join()
        self.close(wait=True, abort=True)

In my case the following happens:

       client1: self.queue.put(None)
       client1: writer_task.join()
       client2: self.queue.put(None)
       client2: writer_task.join()
       client1: self.close(wait=True, abort=True)
             self.server._trigger_event('disconnect')
                    emit('sig-user-leaved', {'client_id': client_id}, room=room, include_self=False)
                         client2: send() { self.queue.put(pkt) } # !!!!!!!! THIS CAUSES THE PROBLEM
       client2: self.close(wait=True, abort=True)                                        

There should be a guard that after 
self.queue.put(None)
no new data is added in the queue before close() is called.

miguelgrinberg:
Thanks. I'll think about how to handle this case, good catch.

Benjaminst123:
@miguelgrinberg This should be fixed by https://github.com/miguelgrinberg/python-engineio/pull/78

miguelgrinberg:
Addressed in release 2.3.1.

PR #78, opened 2018-09-29
Fix file descriptor leakage

Benjaminst123:
Do not wait for the queue to be emptied when the writer thread is terminated.
As the writer thread is no longer active, the queue will never be emptied. This should fix https://github.com/miguelgrinberg/python-engineio/issues/56


miguelgrinberg:
This was missing the asyncio counterpart fix. I have committed this fix along with a few related ones myself. Thanks!

PR #85, opened 2018-12-10
Add function to shutdown the server gracefully.

Viatorus:
Currently, it is not possible to shut down the aiohttp server gracefully, because the _service_task is running in an endless loop.

This PR should fix this problem.


engineio/asyncio_server.py
```diff
@@ -220,24 +231,37 @@ def start_background_task(self, target, *args, **kwargs):
...
+        if not until_shutdown:
+            return await asyncio.sleep(seconds)
+
+        # Sleep a short time and recheck if server is still running.
+        fractional, integer = math.modf(seconds * 10)
+        for i in range(int(integer)):
+            await asyncio.sleep(0.1)
+            if not self.running:
+                return
+        return await asyncio.sleep(fractional)
```
miguelgrinberg: What you are doing here is a somewhat busy wait loop. This is going to add CPU load, so it's not something that I find acceptable. I would actually prefer to leave the sleep function working as a sleep function, and build a better mechanism for the task service to be awaken if it needs to exit, maybe with an `Event` object or similar.

miguelgrinberg:
@Viatorus Thanks! Overall I'm okay with the idea of adding the `shutdown()` method.

The changes to the `sleep()` function that you made are not the way I would do it though, because you rely on polling to determine when it is time to exit and that requires constant CPU work. Let me know if you want to continue working on the PR or not, I can do this myself as well. Thanks again.

Viatorus:
@miguelgrinberg Thanks for your feedback. I updated this part. This change seems to look much cleaner to me. What do you think?

I have never used async/await before.

engineio/asyncio_server.py
```diff
@@ -63,6 +67,13 @@ def attach(self, app, engineio_path='engine.io'):
         engineio_path = engineio_path.strip('/')
         self._async['create_route'](app, self, '/{}/'.format(engineio_path))
 
+    async def shutdown(self):
+        """Shutdown the Engine.IO server."""
+        self.running = False
```
miguelgrinberg: You do not need `self.running` anymore, right?

engineio/asyncio_server.py
```diff
@@ -63,6 +67,13 @@ def attach(self, app, engineio_path='engine.io'):
...
+
+        self.__service_task.cancel()
+        await self.__service_task
```
miguelgrinberg: How long does this take? It would seem you could be waiting up to a full minute until the task exits.

Viatorus:
Thank you for your feedback.

It didn't worked this way, because a task never reaches the state cancelled if it is running. There was also an mistake with: `while self.__service_task.cancelled()`.

Now it should work. The cancel exception raises immediately (also while sleeping). This is now the break condition. No waiting required.

miguelgrinberg:
Thanks, this looks better. I need to test it a bit, and also add a compatible feature for the synchronous server case. As soon as I have that worked out I'll merge and make a new release.

Viatorus:
Great, thank you!

miguelgrinberg:
@Viatorus how are you shutting down aiohttp gracefully? I don't really see what the problem is, when I hit Ctrl-C to stop the server nothing happens, and I need to hit Ctrl-C a second time to get the server to exit. This happens even if I don't start the service task.

Viatorus:
Did you connect to your socket.io first?

miguelgrinberg:
@Viatorus I'm using the `examples/server/aiohttp/latency.py` example in this repository.

PR #90, opened 2019-01-11
Prevent side effect in test_async_modes_wsgi (fixes #89)

### LINKED ISSUE #89: test regresion in 3.2.1 async gevent
I somehow can't make the teset suite pass anymore, failing with the following error:

I have also tried backporting bc128c2e3f41a69d855acabdbdbad072f662df92 but it didn't change the result.

python 3.7
gevent 1.3.7
gevent-websocket 0.10.1
websocket-client 0.54.0
python-urllib3 1.24.1
eventlet 0.24.1

```python
======================================================================
FAIL: test_async_mode_gevent (tests.common.test_server.TestServer)
----------------------------------------------------------------------
  File "/usr/lib/python3.7/unittest/mock.py", line 1195, in patched
    return func(*args, **keywargs)
  File "/build/python-engineio/src/python-engineio-3.2.1/tests/common/test_server.py", line 193, in test_async_mode_gevent
    self.assertEqual(s._async['queue'], 'foo')
AssertionError: <class 'gevent._queue.JoinableQueue'> != 'foo'

======================================================================
FAIL: test_async_mode_gevent_without_websocket (tests.common.test_server.TestServer)
----------------------------------------------------------------------
  File "/usr/lib/python3.7/unittest/mock.py", line 1195, in patched
    return func(*args, **keywargs)
  File "/build/python-engineio/src/python-engineio-3.2.1/tests/common/test_server.py", line 216, in test_async_mode_gevent_without_websocket
    self.assertEqual(s._async['queue'], 'foo')
AssertionError: <class 'gevent._queue.JoinableQueue'> != 'foo'

----------------------------------------------------------------------
Ran 408 tests in 2.832s

FAILED (failures=2)
Test failed: <unittest.runner.TextTestResult run=408 errors=0 failures=2>
error: Test failed: <unittest.runner.TextTestResult run=408 errors=0 failures=2>
```

some help would be appreciated

felixonmars:
The gevent driver was successfully imported here when trying to test async_mode='gevent'. This makes the later mock in test_async_mode_gevent (in test_server.py) fails to actually patch the gevent modules. Removing the imported gevent driver from sys.modules fixes the side effect.

### DIFF
--- a/tests/asyncio/test_asyncio_server.py
+++ b/tests/asyncio/test_asyncio_server.py
@@ -115,6 +115,7 @@ class TestAsyncServer(unittest.TestCase):
                           async_mode='gevent_uwsgi')
         self.assertRaises(ValueError, asyncio_server.AsyncServer,
                           async_mode='threading')
+        sys.modules.pop('engineio.async_drivers.gevent')
 
     @mock.patch('importlib.import_module')
     def test_attach(self, import_module):
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,7 @@ deps=
     urllib3
     websocket-client
     websockets
+    gevent
 basepython =
     flake8: python3.5
     py27: python2.7


tests/asyncio/test_asyncio_server.py
```diff
@@ -115,6 +115,7 @@ def test_async_modes_wsgi(self):
                           async_mode='gevent_uwsgi')
         self.assertRaises(ValueError, asyncio_server.AsyncServer,
                           async_mode='threading')
+        del sys.modules['engineio.async_drivers.gevent']
```
anthraxx: This should be guarded so tests work on native environments when gevent is not available:
```python
        if 'engineio.async_drivers.gevent' in sys.modules
            del sys.modules['engineio.async_drivers.gevent']
```

miguelgrinberg:
So here is what I don't understand. Take a look at this session in the Python console:

```
Python 3.7.1 (default, Nov 17 2018, 18:50:19)
[Clang 10.0.0 (clang-1000.10.44.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.modules['engineio.async_drivers.gevent']
  File "<stdin>", line 1, in <module>
KeyError: 'engineio.async_drivers.gevent'
>>> sys.modules['engineio.async_drivers']
  File "<stdin>", line 1, in <module>
KeyError: 'engineio.async_drivers'

>>> from engineio.async_drivers import gevent
  File "<stdin>", line 1, in <module>
  File "/Users/miguelgrinberg/Documents/dev/flask/python-engineio/engineio/async_drivers/gevent.py", line 3, in <module>
    import gevent
ModuleNotFoundError: No module named 'gevent'

>>> sys.modules['engineio.async_drivers.gevent']
  File "<stdin>", line 1, in <module>
KeyError: 'engineio.async_drivers.gevent'
>>> sys.modules['engineio.async_drivers']
<module 'engineio.async_drivers' from '/Users/miguelgrinberg/Documents/dev/flask/python-engineio/engineio/async_drivers/__init__.py'>
>>> sys.modules['engineio.async_drivers'].gevent
  File "<stdin>", line 1, in <module>
AttributeError: module 'engineio.async_drivers' has no attribute 'gevent'
```

So I don't see what the problem is. The `gevent` package is not installed while the tests run (this was done on purpose long ago, it's not needed for the tests). As you can see importing the gevent driver package does not leave the loaded modules in a bad state, because the lack of gevent prevents it.

anthraxx:
It happens when it is installed, we run the test suite to ensure there are no regressions in the environment it will be used and not a artificial tox environment therefor gevent is installed while we test.

miguelgrinberg:
> in the environment it will be used

There are lots of sensible environments in which this package can be used. In fact most people do not use gevent. So how can you test all the different environments in which this package can be used? Do you also install aiohttp, sanic, tornado, eventlet, uvicorn, gunicorn and uwsgi?

These are unit tests, they are not functional tests. They are not designed to run in a production environment. Really the only problem is that the AsyncServer is importing gevent stuff, which really makes no sense. So having gevent installed affects the way the test behaves. I can look into that, the way you are addressing it is not really what I intended, I don't want to add yet another big package into the test suite.

miguelgrinberg:
Wow. Actually you do install most of those dependencies! Why do you do that, given that these are not "hard" dependencies? That does not make any sense these are all optional, there is no use case in which you need all of them installed together.

anthraxx:
Running the unit test suite is the way everything is tested on a distro level, even if I get your point that it's not permutates we still run test as a rule.

Yes I'm aware of the optional deps already since I digged deeper into this lib to resolve the problem we are facing with the unit test, they will all be optional dependencies is the next release.

Don't add gevent to tox if you don't wish, all we care is the conditional del line so it's not borked when it's available during test run.

miguelgrinberg:
Okay great. Can you give master a try on your environment?

anthraxx:
tests seem to pass now, thanks

miguelgrinberg:
This issue is addressed in release 3.2.3 in a different way.

PR #93, opened 2019-01-25
Set cookies to websocket connection

agavignet:
Hi Miguel,

I am testing https://github.com/miguelgrinberg/python-socketio, it's working well so far but I encountered issues in a production environment.

Particularly when there is a load balancer on top of socket.io server.

To handle websockets, load balancers usually set a cookie in order to keep the client sticky.
Unfortunately cookies are not handled in `python-engineio`, hence this PR.

**Here is a schema to illustrate the issue:**

cookie-engineio

Regarding the fix, all cookies present are forwarded. Another approach could be to have a cookie property in `Client` class, like it's done on server side (https://github.com/miguelgrinberg/python-engineio/blob/master/engineio/server.py#L484) to only take care of one cookie.

**Some readings about the problem:**

https://deepstreamhub.com/blog/load-balancing-websocket-connections/
https://socket.io/docs/using-multiple-nodes/
https://www.haproxy.com/blog/websockets-load-balancing-with-haproxy/

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -202,6 +202,9 @@ class AsyncClient(client.Client):
             await self._receive_packet(pkt)
 
         if 'websocket' in self.upgrades and 'websocket' in self.transports:
+            # apply cookies to websocket connection's headers
+            headers['Cookie'] = r.getheaders().get('Set-Cookie')
+
             # attempt to upgrade to websocket
             if await self._connect_websocket(url, headers, engineio_path):
                 # upgrade to websocket succeeded, we're done here
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -298,6 +298,9 @@ class Client(object):
             self._receive_packet(pkt)
 
         if 'websocket' in self.upgrades and 'websocket' in self.transports:
+            # apply cookies to websocket connection's headers
+            headers['Cookie'] = r.getheaders().get('Set-Cookie')
+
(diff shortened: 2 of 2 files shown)


miguelgrinberg:
I agree with the high-level concept of your solution, I should have thought of this myself.

But your implementation has a few problems:

1. If there is no `Set-Cookie` header, you should not be adding a `Cookie` header to the follow-up request.
2. If there are multiple `Set-Cookie` headers, you should add all of them to the next request.
3. If the `Set-Cookie` header comes with extra options such as expiration, domain, etc. all those options need to be removed and only the cookie name and value should be sent in the follow-up request.

agavignet:
Thanks for your quick feedback! Happy to see this issue is legit to you.
I agree with your points, I will change the implementation that way.

miguelgrinberg:
@agavignet you may want to see if there is something that can be done with the `http.cookiejar` package from the standard library.

miguelgrinberg:
I have implemented this myself in a simpler way. I realized that the requests package does cookie management, replacing urllib3 with it solves this problem. Closing.

agavignet:
@miguelgrinberg I should have keeping inform you on my progress, was just unittesting my fix. But good point is that replacing urllib3 was the best to do (http.cookiejar is not compliant with it btw) and I would not have done that change in your library by myself. So if it's working this way, that's great! Thanks for your time ;)

PR #95, opened 2019-02-10
Fix hang on KeyboardInterrupt when running with asyncio.

iksteen:
The bare except in the _service_task function also catches and
ignores asyncio.CancelledError exceptions. This means the function
will never gracefully exit when cancelling the task.

The solution I propose here adds an additional except handler
specifically for asyncio.CancelledError which will then stop the
service task.

This allows f.e. aiohttp to cancel all pending tasks and allows
the process to terminate even when there are connected clients.

This most likely fixes https://github.com/miguelgrinberg/python-socketio/issues/230


miguelgrinberg:
Thank you so much! Looks like I implemented this for the standard server, but somehow forgot to do it for the asyncio version.

PR #97, opened 2019-02-21
Pass cookies at websocket connection creation

agavignet:
This follows https://github.com/miguelgrinberg/python-engineio/pull/93

Switching `urllib3` to `requests` was a right move but there is still an uncovered case.
Indeed cookies are stored in `Session` object so are well handled for long polling for example, but they are not passed when websocket connection is created.

Implementation:
- Get cookie from `requests.Session()`
- Pass those cookies to `cookie` parameter of `create_connection` and `connect` methods of `websocket` class

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -329,8 +329,11 @@ class Client(object):
             self.logger.info(
                 'Attempting WebSocket connection to ' + websocket_url)
         try:
+            cookies = None
+            if self.http:
+                cookies = '; '.join(["{}={}".format(x.name, x.value) for x in self.http.cookies])
             ws = websocket.create_connection(
-                websocket_url + self._get_url_timestamp(), header=headers)
+                websocket_url + self._get_url_timestamp(), header=headers, cookie=cookies)
         except ConnectionError:
             if upgrade:
                 self.logger.warning(
--- a/tests/common/test_client.py
+++ b/tests/common/test_client.py
@@ -488,7 +488,7 @@ class TestClient(unittest.TestCase):
         self.assertEqual(c.upgrades, [])
         self.assertEqual(c.transport(), 'websocket')
         self.assertEqual(c.ws, create_connection.return_value)
-        create_connection.assert_called_once()
+        self.assertEqual(len(create_connection.call_args_list), 1)
         url = create_connection.call_args[0][0]
         create_connection.assert_called_once_with(
             url, header={}, cookie=None)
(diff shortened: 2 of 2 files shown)


miguelgrinberg:
I haven't looked at this in detail, but honestly I would prefer to keep the completely unrelated logic of dealing with the Cookie header outside of this project. Requests obviously knows how to do this, and so that aiohttp for the asyncio side. My preference would be to let those packages generate the header, if at all possible.

agavignet:
I had a look to websocket lib code and particularly at `create_connection` method. It appears that `cookie` parameter is handled. We can take advantage of this to simplify the fix and decrease impacts. Here is the minimal fix I've come to:

```
diff --git a/engineio/client.py b/engineio/client.py
index 52550e2..6cf7c5a 100644
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -329,8 +329,11 @@ class Client(object):
             self.logger.info(
                 'Attempting WebSocket connection to ' + websocket_url)
         try:
+            cookies = None
+            if self.http:
+                cookies = '; '.join(["{}={}".format(x.name, x.value) for x in self.http.cookies])
             ws = websocket.create_connection(
-                websocket_url + self._get_url_timestamp(), header=headers)
+                websocket_url + self._get_url_timestamp(), header=headers, cookie=cookies)
         except ConnectionError:
             if upgrade:
                 self.logger.warning(
```

What do you think?

miguelgrinberg:
@agavignet yes, that seems like a much better option!

agavignet:
:+1: Then I will push this version of the fix in the next coming days (I need to perform deeper tests and and check unittests) ;)

agavignet:
I still need to:
- fix `assert_called_once` which is apparently not present in python3.5
- test asyncio client on a production environment

agavignet:
@miguelgrinberg The PR is ready for a review ;)

Actually I got troubles to implement the fix in async implementation. 
Indeed it's using websocket**s** lib which unfortunately does not handle `cookie` parameter like websocket lib does.
See `create_connection` of `websockets/client.py`.
```
websockets/client.py", line 440, in __init__
    self._creating_connection = loop.create_connection(factory, host, port, **kwds)
TypeError: create_connection() got an unexpected keyword argument 'cookie'
```
There is actually no reference for `cookie` term in the websocket**s** codebase :/

Anyway at least this PR is fixing the sync Client :)

miguelgrinberg:
👍 This looks good. I can take care of the async case. Thanks!

miguelgrinberg:
This is now merged, and I added the support for the asyncio client. Thank you so much!

PR #109, opened 2019-05-07
Support for Tornado 6

Michotastico:
Hello!

I was working with the library python-socketio and by a problem with the connect I updated it to the latest version (and tornado too) and the following error appears:

```
site-packages/engineio/async_drivers/tornado.py:30: RuntimeWarning: coroutine 'WebSocketHandler.get' was never awaited
  super().get(*args, **kwargs)
```

It happens with Tornado version 6< and block part of the library. And at least in my user-case, only adding `await` in that line the issue was fixed, so I submit this pull request to help with compatibility issues.

### DIFF
--- a/engineio/async_drivers/tornado.py
+++ b/engineio/async_drivers/tornado.py
@@ -27,7 +27,7 @@ def get_tornado_handler(engineio_server):
 
         async def get(self, *args, **kwargs):
             if self.request.headers.get('Upgrade', '').lower() == 'websocket':
-                super().get(*args, **kwargs)
+                await super().get(*args, **kwargs)
             await engineio_server.handle_request(self)
 
         async def post(self, *args, **kwargs):
--- a/engineio/async_drivers/tornado.py
+++ b/engineio/async_drivers/tornado.py
@@ -130,7 +130,7 @@ def make_response(status, headers, payload, environ):
         tornado_handler.write(payload)
         tornado_handler.finish()
     except RuntimeError as error:
-        if not "Method not supported for Web Sockets" in str(error):
+        if "Method not supported for Web Sockets" not in str(error):
             raise
 
 
--- a/engineio/async_drivers/tornado.py
+++ b/engineio/async_drivers/tornado.py
@@ -27,7 +27,7 @@ def get_tornado_handler(engineio_server):
 
         async def get(self, *args, **kwargs):
             if self.request.headers.get('Upgrade', '').lower() == 'websocket':
-                await super().get(*args, **kwargs)
+                await asyncio.coroutine(super().get)(*args, **kwargs)
             await engineio_server.handle_request(self)
 
         async def post(self, *args, **kwargs):
@@ -122,12 +122,16 @@ def make_response(status, headers, payload, environ):
     """This function generates an appropriate response object for this async
     mode.
     """
-    tornado_handler = environ['tornado.handler']
-    tornado_handler.set_status(int(status.split()[0]))
-    for header, value in headers:
-        tornado_handler.set_header(header, value)
-    tornado_handler.write(payload)
-    tornado_handler.finish()
+    try:
+        tornado_handler = environ['tornado.handler']
+        tornado_handler.set_status(int(status.split()[0]))
+        for header, value in headers:
+            tornado_handler.set_header(header, value)
+        tornado_handler.write(payload)
+        tornado_handler.finish()
+    except RuntimeError as error:
+        if not "Method not supported for Web Sockets" in str(error):
+            raise
 
 
 class WebSocket(object):  # pragma: no cover
--- a/engineio/async_drivers/tornado.py
+++ b/engineio/async_drivers/tornado.py
@@ -1,6 +1,7 @@
 import asyncio
 import sys
 from urllib.parse import urlsplit
+from .. import exceptions
 
 try:
     import tornado.web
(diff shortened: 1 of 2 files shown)


Michotastico:
Update. It works fine for connection, but in case of disconnection with the socket a Exception appears and its not handled.

```
  File "/virtualenv/lib/python3.7/site-packages/tornado/web.py", line 1699, in _execute
    result = await result
  File "/virtualenv/lib/python3.7/site-packages/engineio/async_drivers/tornado.py", line 31, in get
    await engineio_server.handle_request(self)
  File "/virtualenv/lib/python3.7/site-packages/engineio/asyncio_server.py", line 209, in handle_request
    packets = await socket.handle_get_request(environ)
  File "/virtualenv/lib/python3.7/site-packages/engineio/asyncio_socket.py", line 89, in handle_get_request
    return await getattr(self, '_upgrade_' + transport)(environ)
  File "/virtualenv/lib/python3.7/site-packages/engineio/asyncio_socket.py", line 128, in _upgrade_websocket
    return await ws(environ)
  File "/virtualenv/lib/python3.7/site-packages/engineio/async_drivers/tornado.py", line 145, in __call__
    await self.handler(self)
  File "/virtualenv/lib/python3.7/site-packages/engineio/asyncio_socket.py", line 148, in _websocket_handler
    data=six.text_type('probe')).encode(always_bytes=False))
  File "/virtualenv/lib/python3.7/site-packages/engineio/async_drivers/tornado.py", line 152, in send
    message, binary=isinstance(message, bytes))
  File "/virtualenv/lib/python3.7/site-packages/tornado/websocket.py", line 339, in write_message
    raise WebSocketClosedError()
tornado.websocket.WebSocketClosedError
```

Any idea how to handle it? With a lower tornado version this not happen (and with python-socketio I could manage the disconnection) and as I see in the tornado code, the change is in the method `write_message`:
```
if self.ws_connection is None:
    raise WebSocketClosedError()
```
to
```
if self.ws_connection is None or self.ws_connection.is_closing():
    raise WebSocketClosedError()	
```

Michotastico:
Finally I managed the errors and the socket is closed in a good way. Also I wrapped the `super().get(*args, **kwargs)` in a coroutine to add support to Tornado < 6 but compatible with Tornado 6 >.

miguelgrinberg:
Isn't this change going to break applications that use Tornado 5?

engineio/asyncio_server.py
```diff
@@ -215,6 +215,13 @@ def __init__(self, server, sid):
                             if sid in self.sockets:  # pragma: no cover
                                 await self.disconnect(sid)
                             r = self._bad_request()
+                        except Exception as error:
```
miguelgrinberg: Do you need to catch all exceptions here? Why not just catch `WebsocketClosedError`?

engineio/async_drivers/tornado.py
```diff
@@ -122,12 +122,16 @@ def make_response(status, headers, payload, environ):
...
-    tornado_handler.write(payload)
-    tornado_handler.finish()
+    try:
+        tornado_handler = environ['tornado.handler']
+        tornado_handler.set_status(int(status.split()[0]))
+        for header, value in headers:
+            tornado_handler.set_header(header, value)
+        tornado_handler.write(payload)
+        tornado_handler.finish()
+    except RuntimeError as error:
```
miguelgrinberg: What is causing a `RuntimeError` here?

Michotastico:
> Isn't this change going to break applications that use Tornado 5?

Using `await asyncio.coroutine(super().get)(*args, **kwargs)` should not break the compatibility with Tornado 5. Because wrapping a coroutine do nothing, and wrapping a normal method allow the use of await.

engineio/asyncio_server.py
Michotastico: I was trying only to catch `WebsocketClosedError` but not add Tornado as a dependency of the library, for that I check if is the correct exception or raise the exception again to be back on the normal flow of execution.

miguelgrinberg: Oh yes, I didn't realize this was the main server, I assumed it was the tornado driver for some reason. No, I definitely don't like this, since all the interactions with Tornado occur from the driver, you should be able to catch this exception there and then raise `EngineIOError`, which this code already knows how to handle.

engineio/async_drivers/tornado.py
Michotastico: Yesterday when I was testing the library with Tornado and Websockets, I saw that the exception is raised when the connection is a websocket (after upgrade) and is closing, Tornado raises the `RuntimeError` using this function. 

```
def _raise_not_supported_for_websockets(*args: Any, **kwargs: Any) -> None:
    raise RuntimeError("Method not supported for Web Sockets")
```

Which is called when the stream is detached. As I could saw, is because we trying to set a method which is not allowed on websocket.
```
def _detach_stream(self) -> IOStream:
        # disable non-WS methods
        for method in [
            "write",
            "redirect",
            "set_header",
            "set_cookie",
            "set_status",
            "flush",
            "finish",
        ]:
            setattr(self, method, _raise_not_supported_for_websockets)
        return self.detach()
```

I did not know if skip `make_response` in case of a websocket will break the compatibility with previous versions, so at least I tried to handle it in the new version.

engineio/asyncio_server.py
Michotastico: Perfect, I will look for the exact location where the exception was raised on the Tornado driver and then raise a `EngineIOError`, I did not realize that in his file Tornado exists as an import. Sorry!

engineio/async_drivers/tornado.py
```diff
@@ -27,7 +28,7 @@ def __init__(self, *args, **kwargs):
 
         async def get(self, *args, **kwargs):
             if self.request.headers.get('Upgrade', '').lower() == 'websocket':
-                super().get(*args, **kwargs)
+                await asyncio.coroutine(super().get)(*args, **kwargs)
```
miguelgrinberg: The coroutine decorator is deprecated, so it's not a good solution here.

miguelgrinberg:
I took your changes and made some improvements. Everything seems to work on both Tornado 5.1.1 and 6.0.2.

PR #113, opened 2019-05-27
Add binary mode support

### LINKED ISSUE #112: Sending binary data using client does not work

akgondber:
While using python-socketio was discovered that emitting data in binary mode does not work in `Client`. Fixes #112.

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -629,7 +629,10 @@ class Client(object):
                 # websocket
                 try:
                     for pkt in packets:
-                        self.ws.send(pkt.encode())
+                        if isinstance(pkt.data, six.binary_type):
+                            self.ws.send(pkt.encode(b64=True))
+                        else:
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Hmm. There shouldn't be a need to encode as base64 with WebSocket, binary packets are supported. The b64 option is for long-polling clients that don't support XHR2.

Can you show me an example of what you are sending that does not work?

akgondber:
Here is a demo: https://github.com/akgondber/python-engineio-issue-112.

miguelgrinberg:
Thank you for the detailed example.

I've found an issue with the websocket client that I'm using, which is sending binary packets as string packets. I coded an identical client using asyncio and that one works fine.

I need to determine if the problem is in how I use the websocket client, or if there is a bug in the client itself, but in any case, this PR is not necessary so I'll close it and keep track of this issue in #112.

Issue #110, opened 2019-05-08
Client does not reconnect after connection timed out


robindboer:
Hi I have a working socketio connection to a remote socketio server. Everything works just fine except at random intervals (mostly after a long period of inactivity) the connection drops out. When the connection drops out I expect it to recover and reconnect to the server automatically, however this does not happen. In the logs I see a recurring pattern when a dropout happens. See the logs below. What is causing this issue and how can I ensure the connection gets picked back up?

```2019-05-08 08:07:23,865 INFO  [socketio.client:402][Thread-21386] Received event "location" [/admin]
2019-05-08 08:07:47,814 INFO  [engineio.client:435][Thread-10977] Sending packet PING data None
2019-05-08 08:08:12,814 INFO  [engineio.client:492][Thread-10977] PONG response has not been received, aborting
2019-05-08 08:08:17,814 ERROR [engineio.client:591][Thread-10978] packet queue is empty, aborting
2019-05-08 08:08:17,817 INFO  [engineio.client:636][Thread-10978] Exiting write loop task
2019-05-08 08:23:32,639 INFO  [engineio.client:556][Thread-10979] Unexpected error "[Errno 110] Connection timed out", aborting
2019-05-08 08:23:32,639 INFO  [engineio.client:564][Thread-10979] Waiting for write loop task to end
2019-05-08 08:23:32,639 INFO  [engineio.client:566][Thread-10979] Waiting for ping loop task to end
2019-05-08 08:23:32,640 INFO  [engineio.client:500][Thread-10977] Exiting ping task
2019-05-08 08:23:32,640 INFO  [engineio.client:576][Thread-10979] Exiting read loop task```

miguelgrinberg:
The reconnects are only triggered when there is a disconnection, not when there is an error.

In this case the client waited the necessary amount of time for a response from the server and didn't receive anything, while the server remained connected, so it appears that the server is not following the Socket.IO protocol.

What kind of server is this? Do you have control over it? Have you configured unusual timeouts for it?

Q5Ca:
Hello @miguelgrinberg, this thread is quite long ago but I have question regarding it. 
My output log is similar to of @robindboer, which the reconnection start after `[Errno 110] Connection timed out` exception. In my case it took 15 minutes average after the timeout wait for PONG response.
After digging the source, I suppose it is because the exception is needed to break the recv loop in `_read_loop_websocket`.
[...]

PR #117, opened 2019-06-07
Fix to reconnection thread not spawned when socket disconnected unexpectly

victormoyanodiaz:
This will fix the socket not being able to reconnect when there was a communication problem (such as bad internet connection or the server delaying the response). 

The reset will be done after calling the disconnected event. 

It will solve #110 and miguelgrinberg/python-socketio#305


miguelgrinberg:
Merged. Thank you!

PR #129, opened 2019-07-16
implement ping and pong events

szero:
Those events are firing in the same places as specified here: 
https://github.com/socketio/engine.io-client#events

I think it is relevant to have these events handled in this project too, not only because implementations in other languages have them but also I found them usefull for implementing custom disconnect handling. (Sometimes I'm in situation where server didn't send close packet, the connection seems to be fine but the ping pong exchange has stopped).

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -136,9 +136,8 @@ class AsyncClient(client.Client):
         :param args: arguments to pass to the function.
         :param kwargs: keyword arguments to pass to the function.
 
-        This function returns an object compatible with the `Thread` class in
-        the Python standard library. The `start()` method on this object is
-        already called by this function.
+        This function returns Future object or Task object wrapped
+        around target function.
 
         Note: this method is a coroutine.
         """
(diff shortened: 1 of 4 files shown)


szero:
Some python3 tests are failing but not on the tests I've added. Something changed in the build environment? https://travis-ci.org/miguelgrinberg/python-engineio/jobs/559657119#L1023

miguelgrinberg:
What is the purpose of this change? My concern is that they add load to the server, since there will be 2 of these every 25 seconds for each client, meaning more than 4 event triggers per client per minute. A server that has 500 clients would be generating over 2000 events a minute, which seems awful lot for something that I suspect almost no application will use.

szero:
I'm requesting this because I'm using engine-io client with a server that have custom protocol over engine-io. Server sends messages with an integer thats incremented on consecutive messages and you have to send the ack message with counter status on ping packet writeout. I know this is quite fringe case, so You can close this PR and I can just go with the fork. Maybe other people who will see this PR will have something to add.

szero:
I just wanted to specify something (wasn't precise in the first post). I've intended this change only for client part of this module and not the server, so there shouldn't be any issues with superfluous server load.

miguelgrinberg:
What's the point of emitting a ping event, since there is nothing you can do in the handler that affects the outcome of the ping/pong dance? Wouldn't it be the same if you set a function to be invoked with a timer at regular intervals that match the ping interval?

miguelgrinberg:
I'm closing this issue, as I don't see it providing a great benefit considering the potential performance implications it adds.

PR #136, opened 2019-09-07
SSL certificate trusting by adding a 'cert' parameter to client connections

ConradBailey:
This allows for the explicit and programmatic trusting of an SSL certificate at connection time. This is important if you're using SSL certificates that are not provided by a widely trusted authority and you want to avoid environment manipulation. Examples could be testing environments or servers that generate their own certificates at run-time.

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -5,6 +5,7 @@ try:
 except ImportError:  # pragma: no cover
     aiohttp = None
 import six
+import ssl
 try:
     import websockets
 except ImportError:  # pragma: no cover
@@ -37,7 +38,7 @@ class AsyncClient(client.Client):
         return True
 
     async def connect(self, url, headers={}, transports=None,
-                      engineio_path='engine.io'):
+                      engineio_path='engine.io', cert=None):
         """Connect to an Engine.IO server.
 
         :param url: The URL of the Engine.IO server. It can include custom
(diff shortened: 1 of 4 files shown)


pasinter:
Any ETA on this?

engineio/asyncio_client.py
```diff
@@ -353,14 +357,15 @@ def _reset(self):
 
     async def _send_request(
             self, method, url, headers=None, body=None,
-            timeout=None):  # pragma: no cover
+            timeout=None, cert=None):  # pragma: no cover
         if self.http is None or self.http.closed:
             self.http = aiohttp.ClientSession()
         http_method = getattr(self.http, method.lower())
+        ssl_context = ssl.create_default_context(cafile=cert)
```
miguelgrinberg: It seems doing this every time a request is sent is not necessary, right? The context can be saved to a `this.ssl_context` member variable and reused.

engineio/asyncio_client.py
```diff
@@ -247,10 +250,11 @@ def _reset(self):
             headers = headers.copy()
             headers['Cookie'] = cookies
 
+        sslopt = {'certfile': cert}
         try:
             ws = await websockets.connect(
                 websocket_url + self._get_url_timestamp(),
-                extra_headers=headers)
+                extra_headers=headers, sslopt=sslopt)
```
miguelgrinberg: Are you sure this is right? My understanding was that this websocket client takes an `SSLContext` object in the `ssl` argument, I don't believe `sslopt` is a valid argument.

miguelgrinberg:
Given that this PR went stale and my questions weren't answered I'm going to close it. I will implement this feature myself when I get a moment.

PR #137, opened 2019-10-01
Added support for SSL connection to unverified host.

sgaron-cse:
This pull request allow devs to set ssl_verify to False when they initialise a python-engineio client which skips SSL certificate validation. This is particularly useful to connect to server protected using self signed certificate. 

Obviously, certificate checking is enabled by default. I made the changes to be the least invasive as possible and since python-socketio lib passes kwargs down to python-engineio. No changes are required in the python-socketio lib to benefit of this feature.

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -1,4 +1,5 @@
 import asyncio
+import ssl
 
 try:
     import aiohttp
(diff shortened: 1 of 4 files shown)


miguelgrinberg:
Thanks. There are a couple of things that need to be done before this can be merged:

1. The tests need to pass, and coverage needs to be 100%
2. the AsyncIO client needs to have the same changes made.

Let me know how much of this work you are willing to do, or if you prefer that I take it to the finish line.

miguelgrinberg:
Thanks! I'll push a release out to pypi today.

PR #138, opened 2019-10-03
avoid loop without yield when count of sockets is too high

Gawen:
For Python 2.x, `sleep_interval` is 0 if `len(self.sockets)` is above
`self.ping_timeout` (default: 60). With eventlet, `self.sleep(0)` looks it does not
yield to the event loop, queueing work which is not processed.  In this
condition, the symptom is the server using 100% of CPU after some time, even
when count of sockets drop below `len(self.sockets)`.

This fix forces `sleep_interval` to be a float, not equal to 0, forcing eventlet
to yield, processing the queued work and avoiding CPU to be at 100%.

This fix does nothing for Python 3.x.

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -652,7 +652,7 @@ class Server(object):
                 continue
 
             # go through the entire client list in a ping interval cycle
-            sleep_interval = self.ping_timeout / len(self.sockets)
+            sleep_interval = float(self.ping_timeout) / len(self.sockets)
 
             try:
                 # iterate over the current clients


miguelgrinberg:
I'll test this. I find it odd that eventlet does not release the CPU with `sleep(0)`, even if it is only Python 2.

maximebf:
I can confirm that this fixes the issue we were having with 100% CPU usage.

Here is our load graph before and after applying the patch:
image

(The load was 100% confirmed to be from the python socketio process)

miguelgrinberg:
So how is this not a bug with eventlet then?

Gawen:
> I find it odd that eventlet does not release the CPU with `sleep(0)`, even if it is only Python 2.

To be fair, this is my interpretation, and this might be incorrect.

> So how is this not a bug with eventlet then?

It is my understanding that `sleep_interval` should be a float. This fixes the 100% CPU loop which never ceases.

miguelgrinberg:
I thought about this and even though I think there is something fishy going on with eventlet, this is a good change, so I'm merging.

PR #140, opened 2019-10-16
Handle case where no original SIGINT handler existed and call signal.…

batesmotel34:
…default_int_handler


miguelgrinberg:
Thanks!

PR #156, opened 2019-12-19
Remove msg limit in aiohttp WebSocket

LARPY:
@miguelgrinberg Aiohttp WebSocketResponse has a default 4 MB limit and we can not change it.
I didn't find ws limits in other places, but here we have a 4 MB limit. It's too low.  I removed 4 MB limit. I think we should remove any size limits from here.


miguelgrinberg:
This limit is to protect the server from a malicious client that sends a large packet in an attempt to make the server crash. I'm not sure what's the best way to change this limit, but disabling it completely is not a good idea.

PR #159, opened 2019-12-26
Fix wrong origin

vasia123:
If the site `domain.com` has a websocket at the address `socket.domain.com`, then without the header `Origin: domain.com`, a connection attempt will throw an error:

```
  File "app.py", line 332, in sync_websockets
    eio.connect(url, transports=['websocket'], engineio_path='socket.io')
  File ".../engineio/client.py", line 184, in connect
    url, headers, engineio_path)
  File ".../engineio/client.py", line 366, in _connect_websocket
    cookie=cookies)
  File ".../websocket/_core.py", line 515, in create_connection
    websock.connect(url, **options)
  File ".../websocket/_core.py", line 226, in connect
    self.handshake_response = handshake(self.sock, *addrs, **options)
  File ".../websocket/_handshake.py", line 81, in handshake
    status, resp = _get_resp_headers(sock)
  File ".../websocket/_handshake.py", line 166, in _get_resp_headers
    raise WebSocketBadStatusException("Handshake status %d %s", status, status_message, resp_headers)
websocket._exceptions.WebSocketBadStatusException: Handshake status 400 Bad Request
```

This commit adds the option `origin` to `client.connect` method and pass it to the websocket.

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -250,11 +250,11 @@ class AsyncClient(client.Client):
                 ssl_context.verify_mode = ssl.CERT_NONE
                 ws = await self.http.ws_connect(
                     websocket_url + self._get_url_timestamp(),
-                    headers=headers, ssl=ssl_context, origin)
+                    headers=headers, ssl=ssl_context, origin=origin)
             else:
                 ws = await self.http.ws_connect(
                     websocket_url + self._get_url_timestamp(),
-                    headers=headers, origin)
+                    headers=headers, origin=origin)
         except (aiohttp.client_exceptions.WSServerHandshakeError,
                 aiohttp.client_exceptions.ServerConnectionError):
             if upgrade:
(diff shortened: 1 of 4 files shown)


miguelgrinberg:
What version of the python-engineio package are you using? This was an issue that was fixed a few weeks ago.

The origin only applies to browsers, other HTTP clients do not need to worry about it.

vasia123:
I am using version 3.11.1 from pip. Python version 3.7.6. The error is reproduced every time upon request. I downloaded the latest version of the library from the git repository. The error continued to appear. Used my patched version of the library without specifying the "origin" - an error appears. Added the origin header to the request - the error is gone.

miguelgrinberg:
Can you share an example application that reproduces this error please?

miguelgrinberg:
I could not reproduce this problem. If you can share example code I can use to test this issue please reopen.

Issue #160, opened 2020-01-15
Couldn't upgrade the protocol in some random cases


dskorykh:
Hello!

I found a confusing issue

As I understand from https://github.com/socketio/engine.io-protocol/blob/master/README.md, when the client sends the `PING` packet, server should respond with the `PONG` packet and then the client sends the `UPGRADE` packet. 
BUT! In some random cases (found only on the iOS browsers) `NOOP` packet which should be sent after `PONG` is processed before `PONG` and client hangs by 25s without sending upgrade packet. 

I found that this issue appears since v3.2.0 release of python-engineio
If I revert these changes https://github.com/miguelgrinberg/python-engineio/commit/919d8ea639d5190d345b22168d6cbfbdebc421af#diff-73af4c5d72e7a60feddc2ce0427079d3R90-R93 to 
just `self.queue.put(pkt)` without packet_backlog, the problem doesn't appear. 

I really couldn't figure out what was causing the problem, how putting messages to the packet_backlog instead of the queue could cause a change in the order of processing low-level packets in the client.

It's reproducible with the simple [Flask-SocketIO example](https://github.com/miguelgrinberg/Flask-SocketIO/blob/master/example/app.py) (async_mode is eventlet with monkey patching)

Debug log from the iOS browser (tested with Safari 13 and Chrome 79.0.3945.73)
```
[Log] socket.io-client:url parse http://192.168.2.247:2000 +0ms (socket.io.js, line 6)
[Log] socket.io-client new io instance for http://192.168.2.247:2000 +0ms (socket.io.js, line 6)
[Log] socket.io-client:manager readyState closed +0ms (socket.io.js, line 6)
[Log] socket.io-client:manager opening http://192.168.2.247:2000 +0ms (socket.io.js, line 6)
[Log] engine.io-client:socket creating transport "polling" +0ms (socket.io.js, line 6)
[Log] engine.io-client:polling polling +0ms (socket.io.js, line 6)
[Log] engine.io-client:polling-xhr xhr poll +0ms (socket.io.js, line 6)
[Log] engine.io-client:polling-xhr xhr open GET: http://192.168.2.247:2000/socket.io/?EIO=3&transport=polling&t=Mz21lBR +0ms (socket.io.js, line 6)
[Log] engine.io-client:polling-xhr xhr data null +1ms (socket.io.js, line 6)
[Log] engine.io-client:socket setting transport polling +3ms (socket.io.js, line 6)
[...]

PR #161, opened 2020-01-22
Do not hang in polling while websocket upgrade is ongoing

salimaboubacar:
Should fix:

- https://github.com/miguelgrinberg/python-engineio/issues/160
- https://github.com/miguelgrinberg/python-socketio/issues/393

And deprecate this other PR which seems to indirectly adresses the issue (probably in a less clean way): https://github.com/miguelgrinberg/python-engineio/pull/76

This fixes a race condition between the client side handling of the PONG packet (websocket upgrade) and the NOOP packet queued for the current long polling connexion. The behavior in the Javascript version of the server is here: https://github.com/socketio/engine.io/blob/cb0ac6fddcad12c454651bf0e1a312a154e228a4/lib/transports/polling.js#L112-L117

### DIFF
--- a/engineio/asyncio_socket.py
+++ b/engineio/asyncio_socket.py
@@ -88,6 +88,8 @@ class AsyncSocket(socket.Socket):
             self.server.logger.info('%s: Received request to upgrade to %s',
                                     self.sid, transport)
             return await getattr(self, '_upgrade_' + transport)(environ)
+        if self.poll_ended:
+            return [packet.Packet(packet.NOOP)]
         try:
             packets = await self.poll()
         except exceptions.QueueEmpty:
@@ -147,6 +149,7 @@ class AsyncSocket(socket.Socket):
             await ws.send(packet.Packet(
                 packet.PONG,
                 data=six.text_type('probe')).encode(always_bytes=False))
+            self.poll_ended = True
             await self.queue.put(packet.Packet(packet.NOOP))  # end poll
 
             try:
--- a/engineio/socket.py
+++ b/engineio/socket.py
@@ -19,6 +19,7 @@ class Socket(object):
         self.connected = False
         self.upgrading = False
         self.upgraded = False
+        self.poll_ended = False
         self.packet_backlog = []
         self.closing = False
         self.closed = False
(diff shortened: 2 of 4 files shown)


miguelgrinberg:
Thanks for looking into this. I like your fix, but I'd like to understand how I can reproduce this problem so that I can then test the fix. Have you been able to reliably reproduce the issue?

salimaboubacar:
You need a race condition to happen in the client. From what I understand in the python client, it can't happen because we do only one GET on polling anyways, and don't try to upgrade while actively polling (right ?), so no race condition is possible with the python client.

The way I reproduced was by spawning thousands of clients in the browser and detecting when one was stuck like this (not so reliable 😅). In the javascript version, we don't send the upgrade packet while there is still a GET polling ongoing: we wait for it to finish first. When a GET finishes we send another one except if we had a pong on the websocket already. The case where we get stuck is if the handler of the GET (receiving the noop) is called before the handler of the pong on websocket: the handler of the GET will send another GET and the handler of the pong will wait for the polling to finish before sending an upgrade packet (this polling will not finish without this patch)

So if you want to simulate what the javascript client can end up doing, the steps are:
1- Send a GET poll asynchronously
2- Probe the websocket, connect, ping and wait for the pong
3- After 1- finishes, send a GET poll (this should return noop immediately but without this patch we are stuck here)
4- Affter 3- finishes, send the upgrade packet on the websocket to finish the upgrade

If you want to reproduce in real life conditions somehow, I suppose you could try to alter the code of the javascript client to force the events to happen in this order.

L312: https://github.com/socketio/engine.io-client/blob/2847411dd03eab4ab003b50d81c7a44f7bead564/lib/socket.js
```javascript
    transport.once('packet', function (msg) {
        ...
    })
```
could become
```javascript
    transport.once('packet', function (msg) {
        setTimeout(() => {
            ...
        }, 1000);
    })
```
(I haven't tested this myself)

miguelgrinberg:
Unfortunately I haven't been able to reproduce this problem, so I cannot test this solution. Do you think you could use the `self.upgrading`  and `self.upgraded` booleans instead of adding the new `self.poll_ended`? You could equate `self.poll_ended == False` to `not self.upgraded or self.upgrading` I think, and `self.poll_ended == True` would be the same as `self.upgraded == True`.

miguelgrinberg:
Superseded by https://github.com/miguelgrinberg/python-engineio/commit/f2cce2bccb7fca58bb6115630d5c221569e52ba4.

PR #169, opened 2020-04-23
ASGI lifespan functions

databasedav:
Added parameters `on_startup` and `on_shutdown` to `ASGIApp` which are functions that are called on receiving their corresponding ASGI event (`'lifespan.startup'` and `'lifespan.shutdown'`). Also updated tests and doc strings. Adapted from [starlette implementation](https://github.com/encode/starlette/blob/5b38e832df5f0b7ec82f3fd04c57d9581c104d36/starlette/routing.py#L508).

### DIFF
--- a/engineio/async_drivers/asgi.py
+++ b/engineio/async_drivers/asgi.py
@@ -1,5 +1,7 @@
 import os
 import sys
+import asyncio
+import traceback
 
 from engineio.static_files import get_static_file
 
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Thanks, overall this looks good, but need to figure out why the tests did not run on your commit...

miguelgrinberg:
Okay, the builds are running, but for some reason GitHub does not show the results. Here is your build: https://travis-ci.org/github/miguelgrinberg/python-engineio/builds/678470805. Just linting issues.

PR #170, opened 2020-04-24
Add ability to set enable_multithread in websocket.create_connection()

agavignet:
As discussed here https://github.com/miguelgrinberg/python-socketio/issues/451

Here is a proposal to handle the issue:

- It does not change current behavior
- Did not add the option in constructor to keep `AsyncClient` and `Client` consistent
- `aiohttp` does not have such option when creating the websocket so no need to implement anything in Async side
- Factorize `create_connection()` calls

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -100,6 +100,7 @@ class Client(object):
         self.queue = None
         self.state = 'disconnected'
         self.ssl_verify = ssl_verify
+        self.ssl_enable_multithread = False
 
         if json is not None:
             packet.Packet.json = json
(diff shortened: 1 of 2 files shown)


agavignet:
@miguelgrinberg forgive me to ping you on this, what's your thought about it?

miguelgrinberg:
I would honestly prefer to see how https://github.com/websocket-client/websocket-client/issues/591 is resolved before I patch it from my side.

miguelgrinberg:
Addressed by always enabling the multithreaded option for now. Will undo if/once the websocket-client package addresses the problem on their side.

agavignet:
Thank you :+1:

PR #181, opened 2020-06-29
Added HTTP proxy support

IsmaelRLG:
This pull request allows developers to establish a proxy when they initialize a python-engineio client. This is particularly useful for connecting to the server when behind a firewall.

I made the changes to make this an optional feature. Although some changes are required in the lib python-socketio to fully benefit from this feature.

I also took the liberty of adding it to the documentation on how to set up a proxy.

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -83,6 +83,20 @@ In the case of the ``asyncio`` client, the method is a coroutine::
 
     await eio.connect('http://localhost:5000')
 
+If the application is behind a firewall or you need to use proxies::
+
+    eio._set_proxy('http://localhost:3128')
+
+Or when the client is initialized::
+
+    import engineio
+
+    # standard Python
+    eio = engineio.Client(proxy='http://localhost:3128')
+
+    # asyncio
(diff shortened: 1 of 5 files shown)


IsmaelRLG:
Hello, any ETA on this enhancement?

IsmaelRLG:
Hello, is there any news if this enhancement will be taken into consideration?

miguelgrinberg:
I have replied to another PR that adds features to the client as follows. The reply applies to this as well.

> Let me see if I can figure out a more generic solution first, because one by one I'm being forced to add arguments that mean nothing to me outside of having to pass them down to another layer. You are adding a cert, then in another open PR another user is adding the proxy arguments.
> 
> My preference would be to accept an already initialized session object, so then this package does not need to care about all these settings directly.

miguelgrinberg:
I have added the `http_session` option to the `Server()` and `AsyncServer()` classes. Can I ask you to test and see if you get proxies passed in a session to work?

miguelgrinberg:
The functionality in this PR is available by passing an already initialized HTTP session. Closing.

PR #182, opened 2020-07-08
Do not catch GeneratorExit and etc.

PaulWasTaken:
Now we catch GeneratorExit here https://github.com/miguelgrinberg/python-engineio/blob/master/engineio/asyncio_server.py#L466 and keep iterating - loop is not closed yet and there is no break/return in the end.
I think we can just catch Exception.

### DIFF
--- a/engineio/asyncio_server.py
+++ b/engineio/asyncio_server.py
@@ -460,10 +460,15 @@ class AsyncServer(server.Server):
                     if not socket.closing and not socket.closed:
                         await socket.check_ping_timeout()
                     await self.sleep(sleep_interval)
-            except (SystemExit, KeyboardInterrupt, asyncio.CancelledError):
+            except (
+                SystemExit,
+                KeyboardInterrupt,
+                asyncio.CancelledError,
+                GeneratorExit,
+            ):
                 self.logger.info('service task canceled')
                 break
-            except Exception:
+            except:
                 if asyncio.get_event_loop().is_closed():
                     self.logger.info('event loop is closed, exiting service '
                                      'task')
--- a/engineio/asyncio_server.py
+++ b/engineio/asyncio_server.py
@@ -463,7 +463,7 @@ class AsyncServer(server.Server):
             except (SystemExit, KeyboardInterrupt, asyncio.CancelledError):
                 self.logger.info('service task canceled')
                 break
-            except:
+            except Exception:
                 if asyncio.get_event_loop().is_closed():
                     self.logger.info('event loop is closed, exiting service '
                                      'task')


miguelgrinberg:
How did you get a GeneratorExit to raise in this function?

PaulWasTaken:
In tests)
Here is some code
*server.py*
```
from aiohttp import web
from socketio import AsyncServer


def create_app() -> web.Application:
    app: web.Application = web.Application(
        middlewares=(
            web.normalize_path_middleware(remove_slash=True, append_slash=False),
        ),
    )

    ws_app: AsyncServer = AsyncServer(
        async_mode='aiohttp',
        cors_allowed_origins=[],
        allow_upgrades=True,
    )
    ws_app.attach(app, socketio_path='/internal/ws')

    @ws_app.event
    async def connect(sid, environ):
        print('connect ', sid)

    @ws_app.event
    async def disconnect(sid):
        print('disconnect ', sid)

    @ws_app.event
    async def testEvent(*args, **kwargs):
        print('ok')

    return app
```

*test_file.py*
```
import pytest
import socketio
from aiohttp.test_utils import TestServer

from server import create_app


@pytest.fixture
async def test_server():
    app = create_app()

    server = TestServer(app, host='127.0.0.1', port=8080)

    await server.start_server()

    yield server

    await server.close()


@pytest.fixture
async def test_client(test_server):
    client = socketio.AsyncClient()

    await client.connect(
        'http://127.0.0.1:8080/',
        transports='websocket',
        socketio_path='internal/ws',
        namespaces='/',
    )

    yield client

    await client.disconnect()


@pytest.mark.asyncio
async def test_smth(test_client):
    await test_client.emit('testEvent')
```

Logs are full with this exception.

miguelgrinberg:
I don't understand why this `GeneratorExit` is raised instead of `CancelledError`, but I see no problem in adding it to the list of watched exceptions.

As a side note, for your tests, you may want to initialize your server with `monitor_clients=False` so that this background task isn't started.

PaulWasTaken:
So your suggestion is adding 
```
except GeneratorExit:
   break
```
block, right?

I mean, what's the point in catching all the exceptions?

Thanks for a quick response, btw

miguelgrinberg:
There is already a break block that catches the exceptions that are supposed to end the task. The `CancelledError` one is there. We can add this one as well.

> I mean, what's the point in catching all the exceptions?

The point is that this task needs to continue to run regardless of what errors occur. If I miss an error then the task will end and then the garbage collection of disconnected clients will not happen anymore.

PaulWasTaken:
I was worried about log message - it may be confusing.
Okay, I will update PR

> The point is that this task needs to continue to run regardless of what errors occur.
I thought here https://github.com/miguelgrinberg/python-engineio/blob/master/engineio/asyncio_server.py#L463 we catch some "specific" exceptions.

miguelgrinberg:
Right, these are all the exceptions that require the task to end. Seems I missed `GeneratorExit`, though I have no idea who/how it is raised. Might be something from pytest...

PR #190, opened 2020-08-26
Allow client to use cert in SSL connection

ventaquil:
Requests library allows to use `cert` param in [`requests.request`](https://requests.readthedocs.io/en/latest/api/#requests.request) so here you have another one param in `Client` contructor to cover this feature.

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -67,6 +67,8 @@ class Client(object):
                        skip SSL certificate verification, allowing
                        connections to servers with self signed certificates.
                        The default is ``True``.
+    :param cert: String path to SSL client cert file (.pem) or tuple with
+                 ('cert', 'key') pair.
     """
     event_names = ['connect', 'disconnect', 'message']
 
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
I believe this is an incomplete implementation, as it does not appear to handle WebSocket connections. It also creates a disparity between the features of the sync and async clients. Finally, there are no unit tests for this feature.

I think in general it is not a maintainable option to keep adding more arguments that are just passed through to the http/websocket clients, so instead I'm considering adding support for passing an already initialized session object.

ventaquil:
> I think in general it is not a maintainable option to keep adding more arguments that are just passed through to the http/websocket clients, so instead I'm considering adding support for passing an already initialized session object.

So should I work on some tests during weekend or abandon this feature?

miguelgrinberg:
Let me see if I can figure out a more generic solution first, because one by one I'm being forced to add arguments that mean nothing to me outside of having to pass them down to another layer. You are adding a cert, then in another open PR another user is adding the proxy arguments.

My preference would be to accept an already initialized session object, so then this package does not need to care about all these settings directly.

miguelgrinberg:
I have added the `http_session` option to the `Server()` and `AsyncServer()` classes. Can I ask you to test and see if you get certs passed in a session to work?

ventaquil:
Of course! I am going to check it on Monday because I am on vacations now. Thanks!

ventaquil:
Thanks @miguelgrinberg, looks fine for me.

PR #192, opened 2020-09-02
Expose ASGI scope in connect environ

### LINKED ISSUE #191: ASGI middleware support
In `engineio.async_drivers.asgi.ASGIApp.translate_request` the ASGI request `scope` is consumed and used to construct an `environ` object, which can be accessed in `connect` event handlers.

However, commonly speaking, [ASGI middleware](https://www.starlette.io/middleware/) such as `SessionMiddleware`, write their information to the ASGI request `scope` object. This is not accessible from the `environ` object and as such I can't benefit from middleware.

It would be very nice if the scope object were exposed to `connect` callbacks via the following simple change in `translate_request`:

```python
    environ = {
        # ...
        'scope': scope,
    }
```

For any future readers facing the same issue, I worked around this problem with an absolutely hideous hack:

```python
import inspect

@sio.event
def connect(sid, environ):
    scope = [f for f in inspect.stack() if f.function == "handle_request" and f.filename.endswith("asyncio_server.py")][-1].frame.f_locals['args'][0]
```

Korijn:
Closes #191

### DIFF
--- a/engineio/async_drivers/asgi.py
+++ b/engineio/async_drivers/asgi.py
@@ -163,6 +163,7 @@ async def translate_request(scope, receive, send):
         'SERVER_PORT': '0',
         'asgi.receive': receive,
         'asgi.send': send,
+        'asgi.scope': scope,
     }
 
     for hdr_name, hdr_value in scope['headers']:


Korijn:
Build failure seems unrelated

miguelgrinberg:
Thank you!

PR #193, opened 2020-09-02
Added options to set cookie SameSite and Secure settings.

tails113:
A fix to allow the user to specify the cookies settings.

### DIFF
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -523,9 +523,9 @@ class Server(object):
             s.connected = True
             headers = None
             if self.cookie:
-                cookie_detail = self.cookie + '=' + sid + '; path=/; SameSite=' + self.cookie_samesite + ';'
+                cookie_detail = self.cookie + '=' + sid + '; path=/; SameSite=' + self.cookie_samesite
                 if self.cookie_secure:
-                    cookie_detail += ' Secure=True;'
+                    cookie_detail += '; Secure=True;'
                 headers = [(
                     'Set-Cookie',
                     cookie_detail
--- a/engineio/server.py
+++ b/engineio/server.py
@@ -48,9 +48,10 @@ class Server(object):
     :param compression_threshold: Only compress messages when their byte size
                                   is greater than this value. The default is
                                   1024 bytes.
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Is this really useful for anything? I'm actually considering deprecating the `cookie` option. I copied it from the JS server many years ago but it doesn't serve any purpose. The JS server is removing it.

cypherlou:
Hey Miguel,
I use the `io` cookie to balance traffic on HAProxy as described in [the socket.io documentation](https://socket.io/docs/using-multiple-nodes/#HAProxy-configuration) (ensuring that any non-upgraded connection or any break in the connection will direct traffic back to the same Origin server) makes me think it might be wise to retain.

Obviously it's pretty straight forward to get HAProxy to set a fresh cookie, which interestingly they appear to do in [the full HAProxy config](https://github.com/socketio/socket.io/blob/master/examples/cluster-haproxy/haproxy/haproxy.cfg) but featuring that approach on the main website suggests to me that quite a number of people would have potentially followed the same methodology I did.

miguelgrinberg:
Just so you are aware, the next major release of the Socket.IO JS server will not write the `io` cookie by default. I went too far when I say they are removing it, they are making it an opt-in option.

How about this, instead of adding two more arguments to the already long list, let's make the cookie argument accept two forms:

- a string, in which case it is the cookie name. Cookie settings are set by default and include `path=/` and `SameSite=Lax`.
- a dict in the form `{'name': 'cookie-name', 'path': '/', 'SameSite': 'Lax'}`. In this case we build the cookie header with the options provided, so you can set the cookie to your liking. This is also future proof, we won't need to add even more arguments if/when new cookie options are needed as the standards evolve.

miguelgrinberg:
Actually, it's going to be three possible ways to set `cookie`. The two in the above comment, plus `None`, in which case the cookie is not set at all.

tails113:
Hi Miguel,
Sounds like a good solution, I have implemented it and added a test. Let me know if there are any issues. Thanks

miguelgrinberg:
Thanks!

PR #198, opened 2020-09-29
WIP: Ensure tests pass on either linux or windows

SeeringPhil:
check_path(...) was raising errors on windows since the returned empty html body does not contain the same newline character depending on the platform.

Since I couldn't see any shared test helpers or config, I advised against creating one just for this variable and just duplicated the code. (If you'd rather extract that to a constant in a shared config / helper file, let me know.)

I ran tox on both linux (3,5 and 3.8) and windows (3.8) and everything seems good to go on my side.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 example/wsgi/static/engine.io.js linguist-vendored
 example/aiohttp/static/engine.io.js linguist-vendored
+
+tests/asyncio/index.html -text
+tests/common/index.html -text
--- a/tests/asyncio/test_async_asgi.py
+++ b/tests/asyncio/test_async_asgi.py
@@ -94,27 +94,31 @@ class AsgiTests(unittest.TestCase):
                 {'type': 'http.response.body', 'body': body.encode('utf-8')}
             )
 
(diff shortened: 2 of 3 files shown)


miguelgrinberg:
The success/failure on this test really depends on how your git is configured to handle line endings, so this fix will just make the test fail for the people that have git configured to not alter line endings, and will make it succeed for people like yourself, who have git configured to convert line endings to CRLF.

A best solution would be to configure this index.html file as not needing line ending conversions in a .gitattributes file. That will make it work for everybody.

SeeringPhil:
That's a good point. I am not familiar with .gitattributes, it's my first time having real issues with CRLF 😂

I gave it a try but it doesn't seem to work yet (on windows). I don't know whether I need to run some git command to re run over those files? I was under the assumption that it would take effect when cloning again. 

I don't have time to continue working on this right now but I might be able to pick it up next week

SeeringPhil:
Misclicked on close 😅😂

miguelgrinberg:
I'm going to close this, as I have added the .gitattributes configuration while I migrated the build to GitHub actions. From now one there's a Linux, Windows and Mac builds running with every commit.

PR #199, opened 2020-09-29
Make sure Async client does not fail to connect on Windows

SeeringPhil:
Add verification that we're not on windows before trying to add a signal handler since this is not supported on that platform.

I'm not quite sure whether or not there would be a better way, but I think this is a pretty quick change. 

I've ran tests on tox, both on linux and on windows and it runs fine ([This other PR](https://github.com/miguelgrinberg/python-engineio/pull/198) of mine proposes a way to fix some "broken" tests on Windows)

The only downside I see is that the coverage dropped from 100% to 99%.

It's getting late so I haven't tried it yet, but adapting this [answer](https://stackoverflow.com/a/35542296) could be promising.
I can give it a try if keeping the perfect score is required to accept the PR 😄

### DIFF
--- a/engineio/asyncio_client.py
+++ b/engineio/asyncio_client.py
@@ -1,7 +1,6 @@
 import asyncio
 import signal
 import ssl
-import sys
 import threading
 
 try:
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
The success/failure on this test really depends on how your git is configured to handle line endings, so this fix will just make the test fail for the people that has git configured to not alter line endings, and will make it succeed for people like yourself, who have git configured to convert line endings to CRLF.

A best solution would be to make this index.html file as not needing line ending conversions in a .gitattributes file. That will make it work for everybody.

SeeringPhil:
I'm not sure to understand you? Is this comment intended for #198 ? 
Because the current (#199) PR does not modify stuff regarding the html file or line endings, but the other one does so I'm a little confused 🤔

miguelgrinberg:
@SeeringPhil sorry, yes, I saw your name and thought this was the other PR. Sorry about that.

miguelgrinberg:
Okay, now commenting on this issue. Can we do this as a "ask for forgiveness" style thing? The call raises `NotImplemented` on all platforms on which signals aren't supported, so there is no need to explicitly single out any platforms. Just make the call and handle the failure. That way, if one day somehow signals start being supported, there is no need to fix the code.

SeeringPhil:
> Okay, now commenting on this issue. Can we do this as a "ask for forgiveness" style thing? The call raises `NotImplemented` on all platforms on which signals aren't supported, so there is no need to explicitly single out any platforms. Just make the call and handle the failure. That way, if one day somehow signals start being supported, there is no need to fix the code.

That's a good design decision, thanks for the tip 🙏

SeeringPhil:
Does that update fit what you had in mind?

PR #201, opened 2020-11-10
Handle non-integer ping interval and timeout

akash-roboticist:
Some socket.io servers (non python) serve `ping_timeout` and `ping_interval` as strings rather than integers, this is causing the client to break at startup.

for reference, here's a response from a server I am connecting to -
```
INFO:engineio.client:WebSocket connection accepted with {'sid': 'some_long_sid', 'upgrades': [], 'pingInterval': '15000', 'pingTimeout': '10000'}
```

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -481,8 +481,8 @@ class Client(object):
                 'WebSocket connection accepted with ' + str(open_packet.data))
             self.sid = open_packet.data['sid']
             self.upgrades = open_packet.data['upgrades']
-            self.ping_interval = open_packet.data['pingInterval'] / 1000.0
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Looks good. Thanks!

akash-roboticist:
@miguelgrinberg thanks for merging.
Since this would unblock my current project, would it be possible to spin out the next release with this fix soon? I'd like to keep using a released version rather than locally modified versions.

Appreciate your work!

miguelgrinberg:
I have two more outstanding bugs that I'd like to address before the next release. In the meantime you can install the master branch direct from git, there is no need to apply local modifications.

akash-roboticist:
Oh yes I forgot about that, let me do that; thanks.

PR #204, opened 2020-11-30
Ignore case when comparing Socket.IO transport param with HTTP_UPGRADE header

komputerwiz:
When proxying websocket connections via `mod_proxy_wstunnel`, Apache sets the HTTP Upgrade header to `Upgrade: WebSocket`. Hence, validation of the `websocket` transport should include this variant, and making the comparison case-insensitive does the trick.

Tests have been added where appropriate to confirm the problem and verify the solution.

### DIFF
--- a/engineio/asyncio_server.py
+++ b/engineio/asyncio_server.py
@@ -244,15 +244,18 @@ class AsyncServer(server.Server):
         elif method == 'GET':
             if sid is None:
                 transport = query.get('transport', ['polling'])[0]
-                if (transport != 'polling' and transport != 'websocket') or \
-                        (transport != 'polling'
-                            and transport != environ.get('HTTP_UPGRADE')):
+                # transport must be one of: 'polling' or 'websocket'.
+                # If 'websocket', the HTTP_UPGRADE header must match.
+                # Some tech likes to use 'WebSocket', so ignore case.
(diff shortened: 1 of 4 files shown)


miguelgrinberg:
I'm okay with the general idea of the change, but I believe the changes that you've made introduce a difference in how this logic works. In the original code, if an upgrade header is not present it is impossible to trick the server into processing a connection. With your changes, I can send `transport=` in the query string and the server will think it is un upgrade to a transport that has an empty string as a name.

Also, advice for this an other PRs that you contribute. It is better to not rearrange the code at the same time you create your PR, unless you have a good reason to. It makes it harder for the reviewer when in addition to the change in question the code has been reorganized.

miguelgrinberg:
Merged after fixing the empty upgrade header. Thanks.

komputerwiz:
Thanks for the quick reply!

You're absolutely right; I was hesitant to rearrange the logic at first for that exact reason. After my first attempt at a revision, I noticed the condition was getting unwieldy: it spanned 5 lines and had multiple nested clauses. I rearranged the logic in an attempt to make the validation conditions clearer and more concise. The condition should have been equivalent, but there might have been a case that I overlooked.

I'm happy I could provide at least some assistance! Thanks for maintaining this library, and keep up the great work!

PR #215, opened 2021-02-11
support for client to verify server with custom ca bundle

bhast:
Adds feature discussed in https://github.com/miguelgrinberg/python-socketio/issues/626

Directly specify path to custom ca bundle using `ssl_verify`

Polling transport already works. This adds support for websocket transport to work as well.

### DIFF
--- a/engineio/client.py
+++ b/engineio/client.py
@@ -65,10 +65,17 @@ class Client(object):
                          when sending requests to the server. Use it if you
                          need to add special client options such as proxy
                          servers, SSL certificates, etc.
-    :param ssl_verify: ``True`` to verify SSL certificates, or ``False`` to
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
The `ssl_verify` argument is old, and I don't really like to continue extended it, since there is a better way to do this now. All this cert related features that really are attributes of the HTTP/WS clients, so the new way to configure your clients is to provide an initialized HTTP session. So in my view the best way to handle this would be to look for a configured CA bundle in the http session and take it from there internally, instead of adding more complexity into the interfaces in this package.

bhast:
Updated PR to look for a configured ca bundle in the http session.

miguelgrinberg:
Merged. Thank you so much!

PR #218, opened 2021-03-11
Remove obsolete 'mock' dependency

mgorny:
The package is using 'unittest.mock' consistently, so remove
the obsolete dependency.


miguelgrinberg:
Thanks. This is a left over from the days when Python 2 was supported.

PR #240, opened 2021-07-04
Remove executable permissions from files that lack shebang lines

musicinmybrain:
A file that does not have a shebang (“#!/usr/bin/python3” or similar)
cannot be executed as a script, so its filesystem permissions should not
include the executable bit.


miguelgrinberg:
Thanks.

Issue #244, opened 2021-09-12
Gevent's async driver is not compatible with the native threading.Thread interface.


dedoussis:
**Describe the bug**
Within the `engineio.async_drivers.gevent` module, the `Thread` implementation claims to be compatible with the standard library's `Thread` class: https://github.com/miguelgrinberg/python-engineio/blob/main/src/engineio/async_drivers/gevent.py#L13

However this does not seem to be the case.
The `engineio.async_drivers.gevent.Thread` class subclasses the Greenlet class instead of the [gevent.threading.Thread](http://www.gevent.org/api/gevent.threading.html#gevent.threading.Thread) class.

**To Reproduce**
```
Python 3.9.0 (default, Apr 22 2021, 08:56:23) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from engineio.async_drivers.gevent import Thread
>>> Thread(lambda: None, daemon=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__() got an unexpected keyword argument 'daemon'
```
**Expected behavior**
Passing the `daemon` kwarg to the Thread instance constructor should not raise a `TypeError`.

dedoussis:
Will be raising a PR to address this, if that's alright

miguelgrinberg:
Why is this an issue? The `Thread` class is used internally. It does not need to fully compatible with Python's `Thread` class.

dedoussis:
Bringing the PR discussion into this issue, so that we have a single thread to focus on.

> As stated in the issue, I don't really see what is the improvement this change brings to the project. Did you have any problems with the current implementation?

I started looking into this implementation because I attempted to use the [flask_socketio.SocketIO.start_background_task](https://github.com/miguelgrinberg/Flask-SocketIO/blob/main/src/flask_socketio/__init__.py#L682) method to start a daemonised thread. This method ultimately calls the [engineio.Server.start_background_task](https://github.com/miguelgrinberg/python-engineio/blob/main/src/engineio/server.py#L445) method which only passes the `target`, `args` and `kwargs` arguments to the constructor that implements the `threading.Thread` interface, and thus it does not allow the construction of a daemonic thread (neither groupped or named). Thus, I attempted to subclass `flask_socketio.SocketIO` and re-implement the `start_background_task` method as follows:

```python
    def start_background_task(
[...]

PR #245, opened 2021-09-12
Make gevent and gevent_uswgi async drivers compatible with threading.…

dedoussis:
…Thread interface

Fixes: https://github.com/miguelgrinberg/python-engineio/issues/244

### DIFF
--- a/src/engineio/async_drivers/gevent.py
+++ b/src/engineio/async_drivers/gevent.py
@@ -3,23 +3,12 @@ from __future__ import absolute_import
 import gevent
 from gevent import queue
 from gevent.event import Event
+from gevent.threading import Thread
 try:
     import geventwebsocket  # noqa
     _websocket_available = True
 except ImportError:
     _websocket_available = False
-
-
-class Thread(gevent.Greenlet):  # pragma: no cover
-    """
-    This wrapper class provides gevent Greenlet interface that is compatible
-    with the standard library's Thread class.
-    """
-    def __init__(self, target, args=[], kwargs={}):
-        super(Thread, self).__init__(target, *args, **kwargs)
-
-    def _run(self):
-        return self.run()
 
 
 class WebSocketWSGI(object):  # pragma: no cover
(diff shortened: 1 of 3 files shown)


src/engineio/async_drivers/gevent.py
```diff
@@ -3,25 +3,14 @@
...
 
-class Thread(gevent.Greenlet):  # pragma: no cover
-    """
-    This wrapper class provides gevent Greenlet interface that is compatible
-    with the standard library's Thread class.
-    """
-    def __init__(self, target, args=[], kwargs={}):
-        super(Thread, self).__init__(target, *args, **kwargs)
-
-    def _run(self):
```
dedoussis, line 21: This is not part of the `threading.Thread` interface.

miguelgrinberg: And why is this a problem? The `Thread` class in this driver is used internally.

dedoussis: I see. I couldn't find any occurrences of this method being used anywhere in the codebase and since this is a private method, I decided to not include it anymore. If that's not the case, I'm more than happy to put it back 👍

miguelgrinberg: Sorry, my question wasn't directly related to the `_run()` method, I was wondering about the whole change. I don't see any benefit in switching to the `Thread` class instead of using the `Greenlet`  class as I'm doing here. There's nothing wrong with with the `Thread` class as far as I know, but my solution based on `Greenlet` has years of testing, so I'm not sure it makes sense to switch to a different implementation without a good reason.

But in any case, you asked about the `_run()` method. This is where you put the code that runs in the greenlet. You can see how this works in [the documentation](https://www.gevent.org/api/gevent.greenlet.html#gevent.Greenlet.__init__).

miguelgrinberg:
As stated in the issue, I don't really see what is the improvement this change brings to the project. Did you have any problems with the current implementation?

src/engineio/async_drivers/gevent.py
dedoussis: Thanks for the `_run()` clarification, that makes complete sense then. I will include it in the class, if we decide to go forward with this PR.

miguelgrinberg:
I will not be merging this change, as discussed in https://github.com/miguelgrinberg/python-engineio/issues/244.

Issue #133, opened 2019-08-06
Provide back-pressure when server is overloaded


ericvanbuhler:
I'm using `python-engineio` in a real-time scenario on an embedded device with fluctuating network connection quality, and I'm seeing that the server can become overloaded, leading to a significant lag that the server doesn't recover from. One possible solution would be to limit the server queue size, but that knob isn't exposed at the module level. Is there another way to provide server load state to the higher layers? I wanted to double check before going about implementing those changes.

miguelgrinberg:
> One possible solution would be to limit the server queue size

Can you clarify what you mean here? This package is not a web server, it relies on an external web server such as gunicorn, etc. What is it exactly that you want to limit?

ericvanbuhler:
Thanks for taking a look, Miguel. I was referring to the queue created here:
https://github.com/miguelgrinberg/python-engineio/blob/441a323349a9473998597f785487984e8f14e73f/engineio/socket.py#L17

It seems that by setting that queue to a limited size, I could provide back-pressure to the higher layers when the server is backed up. I'm using eventlet in my case.

miguelgrinberg:
@ericvanbuhler so then you mean "queues' in plural. There's one of those per client.

Are you thinking in returning an error when you try to emit an event and the queue has reached certain size? Note that this is half of the equation, since you also have the incoming events.

A more end-to-end approach to control slow downs would be to use callbacks for each emitted event. The backend application can emit an event, and wait until it gets the acknowledgement callback to emit again. This would account for the whole round trip, I think.

ericvanbuhler:
I was thinking more along the lines of blocking when I try to emit an event and the queue has reached a certain size. That way engineio would not drop any events, but the app could drop messages while it waits for the server to catch up. An error would also work, but it would probably require more handling internal to engineio to ensure control message are not dropped, right?

In my case most of the traffic is from the server to the clients, so I hadn't thought much about traffic coming from the clients. You're right that this would not a complete solution.

I considered using callbacks, but I was concerned about the following:
- Complexity of managing individual clients: my app is broadcasting to all clients, and ensuring that all clients have received the packet before sending the next would add some complexity.
- The overhead of ACKing each packet: this could have an impact on the real-time performance of my app.

miguelgrinberg:
> I was thinking more along the lines of blocking when I try to emit an event and the queue has reached a certain size.

[...]

PR #249, opened 2021-09-19
Adding basic back pressure ability

jackmead515:
Adds a back_pressure argument to the Socket.py send() function. This is an argument that has to be >= 1.

I fully understand that it isn't as fleshed out as it could be. So if requests are made to add more testing, better linting practices, etc, let me know. I think this would be a great addition to the lib.

### DIFF
--- a/src/engineio/asyncio_socket.py
+++ b/src/engineio/asyncio_socket.py
@@ -66,12 +66,32 @@ class AsyncSocket(socket.Socket):
             return False
         return True
 
-    async def send(self, pkt):
-        """Send a packet to the client."""
+    async def send(self, pkt, apply_bp):
+        """
+        Send a packet to the client.
+
+        params:
+            - apply_bp - applies backpressure handling by limiting
+                         the amount of messages that can go into the
+                         queue. This should not be applied to control
+                         messages.
+        """
         if not await self.check_ping_timeout():
             return
-        else:
-            await self.queue.put(pkt)
+        
+        # back pressure enabled. Check if need to limit
+        if apply_bp and self.queue.qsize() >= self.server.back_pressure_size:
+            self.server.logger.warning(
+                    '%s: back pressure applied. Not Sending packet %s data %s',
+                    self.sid, packet.packet_names[packet.packet_type],
+                    packet.data if not isinstance(packet.data, bytes)
+                    else '<binary>')
+            return
+
+        self._send(pkt)
+
+    async def _send(self, pkt):
+        await self.queue.put(pkt)
         self.server.logger.info('%s: Sending packet %s data %s',
                                 self.sid, packet.packet_names[pkt.packet_type],
                                 pkt.data if not isinstance(pkt.data, bytes)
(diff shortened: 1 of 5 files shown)


src/engineio/asyncio_socket.py
```diff
@@ -66,12 +66,32 @@ async def check_ping_timeout(self):
...
-            await self.queue.put(pkt)
+        
+        # back pressure enabled. Check if need to limit
+        if apply_bp and self.queue.qsize() >= self.server.back_pressure_size:
+            self.server.logger.warning(
+                    '%s: back pressure applied. Not Sending packet %s data %s',
+                    self.sid, packet.packet_names[packet.packet_type],
+                    packet.data if not isinstance(packet.data, bytes)
+                    else '<binary>')
+            return
```
miguelgrinberg, line 89: This is just dropping packets, without giving the application any way to know. If the app is expecting a callback, that'll never arrive. I think this makes it impossible to write a robust application.

A more reasonable approach is what I proposed in https://github.com/miguelgrinberg/python-engineio/issues/133#, which is to set a fixed queue size and let the queue block the app if it reaches the maximum size. With this, no packets are lost, and the service is effectively able to tell the app to slow down by blocking it.

jackmead515: I tried to design in such a way that only packets that the server is emitting to the client can be dropped. All others, pings and what not, are fine and are pushed into the queue (at least this is the intention).

There are definitely use cases where you would want to drop the packets. In my use case, I'm sending video to a user. I don't care about a callback as the overhead is too much for me and I don't want the server to buffer all these video packets if the client is unable to keep up with the server. This will cause delayed playback. I think it would be better to only drop packets that are the data itself (not control packets as in heartbeats). That's why I defined a methodology checking the qsize() but not actually using a blocking method to deadlock the queue when I call emit(). deadlocking that queue would be dangerous indeed and I'd love to see a way to actually handle that.

Perhaps the terminology to use here is: "drop_packet_limit". But, not disagreeing that a fixed queue size implementation to prevent dropping off packets would be nice.

miguelgrinberg: I get that this works for your case, but it is an extremely limited solution. I still don't understand how control messages are let through. I see that Engine.IO control message are allowed, but what happens when the Socket.IO layer calls the `send()` method in Engine.IO? Socket.IO also has control messages such as callbacks, errors, open/close messages. All those are at risk of being dropped as well and that breaks the protocol.

I suggest you look at how the JS Socket.IO implements "volatile"` messages. I would be okay with an implementation that matches of volatile that matches or is close to the reference JS project.

jackmead515: Sure I'll have to look into exactly how they define it. If that's what you'd like to see in the project.

I see this more as an advanced option that can only be toggled on in a senario like I'm discussing. Basically, since TCP is the way that it is, I want it to work more closely to a UDP pipe in which it's fine if a packet or two doesn't show up if the client doesn't have the bandwidth. Video data doesn't care bout that.

But cool sounds too complicated to be accomplished in a few evenings after work with this project.

PR #250, opened 2021-09-28
Remove user input from Bad Request responses

RedRoserade:
As discussed via e-mail, this PR removes user-input from the HTTP Bad Request responses that result from invalid transports or session IDs.

All unit tests seem to pass locally.


miguelgrinberg:
Thank you so much!

PR #253, opened 2021-10-14
fix gevent_uwsgi, binary encode

ForJobsGit:
fix gevent_uwsgi
fix binary encode


miguelgrinberg:
This is from the Engine.IO protocol specification: "For binary data the packet type is not included, since only "message" packet type can include binary.". So your second fix is invalid.

The first fix I still don't understand.

Issue #251, opened 2021-10-07
Incompatible with Sanic 21.9.1


wtfuzz:
**Describe the bug**
It seems engineio is incompatible with Sanic 21.9.1

```
[2021-10-06 18:04:25 -0700] [349517] [ERROR] Exception occurred while handling uri: 'http://localhost:8080/socket.io/?transport=polling&EIO=4&t=1633568665.1407533'
Traceback (most recent call last):
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/signals.py", line 73, in get
    group, param_basket = self.find_route(
  File "", line 68, in find_route
sanic_routing.exceptions.NotFound: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/http.py", line 141, in http1
    await self.protocol.request_handler(self.request)
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic_plugin_toolkit/realm.py", line 550, in _handle_request_21_03
    _ = await real_handle(request)
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/app.py", line 769, in handle_request
    await self.dispatch(
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/signals.py", line 163, in dispatch
    return await dispatch
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/signals.py", line 114, in _dispatch
    raise e
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/signals.py", line 111, in _dispatch
    group, handlers, params = self.get(event, condition=condition)
  File "/home/fuzz/silvertip/silvertip-venv/lib/python3.8/site-packages/sanic/signals.py", line 86, in get
    raise NotFound(message % tuple(terms))
sanic_routing.exceptions.NotFound: Could not find signal http.lifecycle.handle
```

**To Reproduce**
Upgrade Sanic to 21.9.1 and attempt to connect a SocketIO/EngineIO client

**Expected behavior**
Successful connection

cnicodeme:
Note that on my case, this yield another error:

> Traceback (most recent call last):
[...]

PR #265, opened 2022-01-03
Fix imports for `sanic>=21.9.0`

13g10n:
# Fix imports for `sanic>=21.9.0`

## Description
There was a change in imports in `sanic` from version 21.9.0, so we need to use `sanic.server.protocols.websocket_protocol.WebSocketProtocol` instead of `sanic.websocket.WebSocketProtocol` as for versions before `21.6.x` ([link to tags diff](https://github.com/sanic-org/sanic/compare/v21.6.2...v21.9.0#diff-17e2193a3741672b54b792ef620715f1d792a2268a573d78eb2dd757f38c94d8R77))

## Changes
+ Add `sanic` version check
+ Update import for `sanic>=21.9.0` to be `sanic.server.protocols.websocket_protocol.WebSocketProtocol`
+ Update sanic example code to latest LTS

### DIFF
--- a/examples/server/sanic/requirements.txt
+++ b/examples/server/sanic/requirements.txt
@@ -1,7 +1,8 @@
-aiofiles==0.3.0
-httptools==0.0.9
-python_engineio
-sanic==0.3.1
-six==1.10.0
-ujson==1.35
-uvloop==0.8.0
+aiofiles==0.8.0
+httptools==0.3.0
+multidict==5.2.0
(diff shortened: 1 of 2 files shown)


13g10n:
Referencing #251

miguelgrinberg:
Thanks! I merged after making a small change in the version detection logic, which I prefer to be in EAFP style, as done to detect when sanic isn't installed.

PR #270, opened 2022-02-08
Do not decode empty text packets

### LINKED ISSUE #269: Receiving empty non-binary packets during connection establishment result in a crash
**Describe the bug**

Receiving empty non-binary packets during connection establishment result in a crash. This is due to a `Packet` with an `encoded_packet = ''` results in an uncaught `IndexError`.

```python
  File "/usr/local/lib/python3.9/site-packages/socketio/client.py", line 660, in _handle_reconnect
    self.connect(self.connection_url,
  File "/usr/local/lib/python3.9/site-packages/socketio/client.py", line 331, in connect
    self.eio.connect(real_url, headers=real_headers,
  File "/usr/local/lib/python3.9/site-packages/engineio/client.py", line 185, in connect
    return getattr(self, '_connect_' + self.transports[0])(
  File "/usr/local/lib/python3.9/site-packages/engineio/client.py", line 324, in _connect_polling
    if self._connect_websocket(url, headers, engineio_path):
  File "/usr/local/lib/python3.9/site-packages/engineio/client.py", line 444, in _connect_websocket
    pkt = packet.Packet(encoded_packet=p)
  File "/usr/local/lib/python3.9/site-packages/engineio/packet.py", line 27, in __init__
    self.decode(encoded_packet)
  File "/usr/local/lib/python3.9/site-packages/engineio/packet.py", line 51, in decode
    b64 = not self.binary and encoded_packet[0] == 'b'
IndexError: string index out of range
```

**To Reproduce**

When receiving an empty packet which is not bytes (e.g. `''`), the Packet class tries to decode this (it did not previous the mentioned commit) and throws an exception as it cannot read the first byte (which is supposed to contain the packet type).

This behavior can occur when establishing a WebSocket connection. In our case, it appeared after the PING packet has been sent and a PONG packet is expected to be received.

https://github.com/miguelgrinberg/python-engineio/blob/06480be268852ffc96793ef52213ad598b85fa69/src/engineio/client.py#L437-L444

Somehow, the packet got not received properly and the `websocket-clients` library `recv` function might return an empty string (see [source](https://github.com/websocket-client/websocket-client/blob/d7aa5c9c244e227041764c63949d3121f84145ac/websocket/_core.py#L353-L368)).
In that case, the empty packet (`''`) is decoded which causes the exception to be raised.

**Expected behavior**

An empty non-binary `encoded_packet` is not decoded and no exception is raised.

**Additional information**

The commit bcd1a42f86aad12460b6eb836a5c317db55cba77 introduced the feature to support binary packets with zero length. 
However, the change introduced this bug here.

sirkrypt0:
The commit bcd1a42f86aad12460b6eb836a5c317db55cba77 introduced the
feature to support binary packets with zero length. However, the
change introduced a bug which is fixed with this commit.

When receiving an empty packet which is not bytes (e.g. ''), the Packet class
tries to decode this (it did not previous the mentioned commit) and
throws an exception as it cannot read the first byte (which is supposed to
contain the packet type).

This behavior can occur when establishing a WebSocket connection. After the
PING packet has been sent (client.py#431), a PONG packet is expected to be
received (client.py#438). However, the WebSocket receive function can
return an empty string which now makes the packet decoding crash.
(see https://github.com/websocket-client/websocket-client/blob/d7aa5c9c244e227041764c63949d3121f84145ac/websocket/_core.py#L353-L368)
As this exception is not caught, it prevents further connection attempts.

Fixes #269

### DIFF
--- a/src/engineio/packet.py
+++ b/src/engineio/packet.py
@@ -23,7 +23,10 @@ class Packet(object):
             self.binary = False
         if self.binary and self.packet_type != MESSAGE:
             raise ValueError('Binary packets can only be of type MESSAGE')
-        if encoded_packet is not None:
+
+        is_binary_encoded_packet = isinstance(encoded_packet, binary_types)
+        if (is_binary_encoded_packet and encoded_packet is not None) or \
+                encoded_packet:
             self.decode(encoded_packet)
 
     def encode(self, b64=False):
--- a/tests/common/test_packet.py
+++ b/tests/common/test_packet.py
@@ -41,6 +41,12 @@ class TestPacket(unittest.TestCase):
         pkt = packet.Packet(encoded_packet=b'4')
         assert pkt.encode() == b'4'
 
+    def test_decode_empty_packet(self):
+        pkt = packet.Packet(encoded_packet='')
+        assert pkt.data is None
+        assert not pkt.binary
+        assert pkt.encode() == '6'
+
     def test_encode_binary_packet(self):
         pkt = packet.Packet(packet.MESSAGE, data=b'\x01\x02\x03')
         assert pkt.packet_type == packet.MESSAGE
--- a/tests/common/test_packet.py
+++ b/tests/common/test_packet.py
@@ -26,8 +26,9 @@ class TestPacket(unittest.TestCase):
         assert pkt.encode() == '4text'
 
     def test_decode_text_packet(self):
-        pkt = packet.Packet(encoded_packet=b'4text')
-        assert pkt.encode() == b'4text'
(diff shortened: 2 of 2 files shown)


sirkrypt0:
@miguelgrinberg While working out a fix for #269 I found that two tests were not doing what they are supposed to, so I fixed that in 1d4548287a83ed0094f63a2a0b26d32413483f39.

miguelgrinberg:
Given that these empty packets are invalid, I have implemented a fix that properly raises an error and halts the connection. Ignoring the empty packet as done in this PR does not seem to me like a good solution.

Issue #264, opened 2021-12-27
Regex CORs


hmajid2301:
**Is your feature request related to a problem? Please describe.**

Would it be possible to add regex CORs as an option, so we can support wildcard domains ? 

**Describe the solution you'd like**

Perhaps you could specify another argument, something like so ? 

`socketio.AsyncServer(async_mode="asgi", regex_cors_allowed_origins=["http://localhost*:3000"])`

**Describe alternatives you've considered**

**Logs**

**Additional context**

miguelgrinberg:
The CORS specification does not support origins given as regular expressions.

hmajid2301:
Yeh sure, I know its not the same but FastAPI have their own implementation to support RegexCORs is there any appetite for this ? Or is there another way I can support wildcard domains.

Thanks

miguelgrinberg:
How about this. Let's say the `cors_allowed_origins` option is extended to support a callable in addition to a string. When using this option, your function will be called to decide if an origin is allowed or not.

hmajid2301:
yeh that sounds like a nice elegant "workaround". Thanks Miguel

PR #273, opened 2022-04-20
feat: allow cors_allowed_origins to be callable

hmajid2301:
Following this conversation https://github.com/miguelgrinberg/python-engineio/issues/264

> How about this. Let's say the cors_allowed_origins option is extended to support a callable in addition to a string. When using this option, your function will be called to decide if an origin is allowed or not. Inside your function you can use anything you want to determine if you allow the origin, including a regex match.

I'm not sure the best way to do this but here goes; check if the `cors_allowed_origins` if it is call it passing in origin. Expect the callable to return a boolean, if true allow request to go through else reject it.

### DIFF
--- a/src/engineio/asyncio_server.py
+++ b/src/engineio/asyncio_server.py
@@ -200,22 +200,19 @@ class AsyncServer(server.Server):
             origin = environ.get('HTTP_ORIGIN')
             if origin:
                 if callable(self.cors_allowed_origins):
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Unfortunately this solution is incomplete, as you haven't done anything to generate a correct CORS response when a callable is used. I have committed a correct implementation to the main branch. Please give it a try and let me know if it works for you.

PR #286, opened 2022-07-19
added timeout to create_websocket()

jpfarias:
Currently the create_websocket() call doesn't have any timeout which means it can wait forever for the websocket to connect.

I made it use the same `request_timeout` as the one used for pooling connection.

This is a pretty small patch, hope it won't be a problem to merge.

### DIFF
--- a/src/engineio/client.py
+++ b/src/engineio/client.py
@@ -417,7 +417,8 @@ class Client(object):
         try:
             ws = websocket.create_connection(
                 websocket_url + self._get_url_timestamp(), header=headers,
-                cookie=cookies, enable_multithread=True, **extra_options)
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Would it be possible to implement the same change on the async client? And adding a test that ensures that the request timeout is used when it is not set to the default? Thanks!

jpfarias:
Yeah, will do! Will update the PR when it is done.

jpfarias:
Sorry for the many commits, I am new to contributing and wasn't aware of all the checks. Hopefully everything is ok now. Thanks again!

miguelgrinberg:
Thanks!

PR #293, opened 2022-10-25
Allow configuring underlying aiohttp websocket

bruce-y:
This PR naively attempts to address https://github.com/miguelgrinberg/python-socketio/issues/486. I am currently running into the same issue and need to be able to configure the `max_msg_size` when calling `ws_connect`.

### DIFF
--- a/src/engineio/asyncio_client.py
+++ b/src/engineio/asyncio_client.py
@@ -66,13 +66,6 @@ class AsyncClient(client.Client):
                               arguments passed to aiohttp.ws_connect().
     """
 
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
While this, in principle, is okay, I do not like the idea of introducing features that are partially implemented. In this case, you have implemented the async client, but there's nothing to support the sync client and the sync and async servers.

bruce-y:
I've added implementation for the sync client as well. I'm not sure what you're envisioning for exposing similar options for the server/async server.

miguelgrinberg:
Merged with some minor changes and added tests. Thanks!

Issue #296, opened 2022-11-23
Resource leaks when using uWSGI WebSockets


juneoh:
**IMPORTANT**: If you have a question, or you are not sure if you have found a bug in this package, then you are in the wrong place. Hit back in your web browser, and then open a GitHub Discussion instead. Likewise, if you are unable to provide the information requested below, open a discussion to troubleshoot your issue.

**Describe the bug**
A WebSocket PONG miss while using uWSGI will leave an open TCP connection on python-engineio, gradually consuming more CPU.

**To Reproduce**
1. Run a python-engineio based websocket server, e.g. using Flask-SocketIO
2. Connect to the server with any websocket client.
3. On the client side, intentionally miss a WebSocket PONG.
4. uWSGI will complain that the pong has been missed, emitting a log ([uWSGI code](https://github.com/unbit/uwsgi/blob/ae489f0881fb358347aabc168a28a96672b0ce3c/core/websockets.c#L70)), resulting `uwsgi_websocket_recv()` to return an empty buffer.
5. Upon the empty buffer, uWSGI's Python binding will raise an IOError during `uwsgi.websocket_recv()` ([uWSGI code](https://github.com/unbit/uwsgi/blob/ae489f0881fb358347aabc168a28a96672b0ce3c/plugins/python/uwsgi_pymodule.c#L1238-L1242)).
6. python-engineio's `uWSGIWebSocket.wait()` catches the IOError and returns None ([python-engineio code](https://github.com/miguelgrinberg/python-engineio/blob/6e93a2c19655b1a2f2dc4af858452505740eac3b/src/engineio/async_drivers/gevent_uwsgi.py#L115-L118)).
7. python-engineio's `Socket._websocket_handler()` will receive the returned None, break out of the main loop ([python-engineio code](https://github.com/miguelgrinberg/python-engineio/blob/6e93a2c19655b1a2f2dc4af858452505740eac3b/src/engineio/socket.py#L229-L241)), join the `writer_task()`, and call `Socket.close()`.
8. `Socket.close()` will attempt to send the CLOSE packet, but since `writer_task()` is already joined, the packet will only be queued, never sent. 
9. Since `uwsgi_disconnect()` is never called in this flow, uWSGI's socket is never explicitly closed, hence leaked.

**Expected behavior**
The connection should be explicitly closed.

**Additional context**
[...]

PR #301, opened 2022-12-09
Catch IOErrors from uWSGI and explicitly close the driver

juneoh:
Fixes https://github.com/miguelgrinberg/python-engineio/issues/296.

- Adds explicit paths toward `uwsgi.disconnect()` to clean up for disconnections during send and receive
- Reorders uWSGIWebSocket.close() so that the greenlet wouldn't step on a closed connection and log an OSError


miguelgrinberg:
Thank you!

PR #303, opened 2023-01-05
Tell the user about versions on version mismatch

iTrooz:
I think telling more info would be a good thing, to help users debug what is going on in these cases


miguelgrinberg:
Thanks. Unfortunately, I'm not going to accept this change. I understand why this may seem like an improvement, but in general it is considered a bad practice for the server to return information about itself in error responses to clients.

Additionally, the engine.io protocol version is a fairly obscure number that very few people understand. Most clients and server do not even mention which protocol version they support except maybe in documentation.

Overall it is better to inform about the version mismatch and then let the developer figure out the correct versions of packages (not of protocols) using the compatibility table in the documentation.

Issue #292, opened 2022-10-19
Object of type 'int' has no len()  error when waiting for messages


wescleveland:
Possibly unhandled incorrect / unexpected response from server to socketio client:

This occurred while debugging some error logic with a socketio client talking to a 3rd party server that was having connection issues.

**Stacktrace**
```object of type 'int' has no len()
Traceback (most recent call last):
  File "/home/ubuntu/code.py", line 313, in connect
    await self.sio.wait()
  File "/lib/python3.10/site-packages/socketio/asyncio_client.py", line 181, in wait
    await self._reconnect_task
  File "/lib/python3.10/site-packages/socketio/asyncio_client.py", line 477, in _handle_reconnect
    await self.connect(self.connection_url,
  File "/lib/python3.10/site-packages/socketio/asyncio_client.py", line 142, in connect
    await self.eio.connect(real_url, headers=real_headers,
  File "/lib/python3.10/site-packages/engineio/asyncio_client.py", line 114, in connect
    return await getattr(self, '_connect_' + self.transports[0])(
  File "/lib/python3.10/site-packages/engineio/asyncio_client.py", line 260, in _connect_polling
    if await self._connect_websocket(url, headers, engineio_path):
  File "/lib/python3.10/site-packages/engineio/asyncio_client.py", line 338, in _connect_websocket
    pkt = packet.Packet(encoded_packet=p)
  File "/lib/python3.10/site-packages/engineio/packet.py", line 27, in __init__
    self.decode(encoded_packet)
  File "/lib/python3.10/site-packages/engineio/packet.py", line 51, in decode
    if not self.binary and len(encoded_packet) == 0:
TypeError: object of type 'int' has no len()
```

This might be an issue with the server sending some strange response to the client, but thought I would submit an issue just in case it is something to look at.  Thanks!

miguelgrinberg:
Did you have logs enabled when this happened? It would help to see the logs leading to this error. Thanks.

wescleveland:
I turned on socketio io logging and engineio logging, below is everything I could fine before / after when it occurs.  The server is still flapping up and down, so let me know if you would like me to test something else.

```Polling connection accepted with {'sid': '321SPblbOnKgEaptAAFX', 'upgrades': ['websocket'], 'pingInterval': 25000, 'pingTimeout': 20000, 'maxPayload': 1000000}
Engine.IO connection established
Sending packet MESSAGE data 0{}
[...]

PR #308, opened 2023-02-08
Potential fixes for two issues

wescleveland:
Hi @miguelgrinberg 

We previously talked in this issue https://github.com/miguelgrinberg/python-engineio/issues/292

Your changes to main fixed that error but I was still encountering two issues.  This PR fixes them _for me_ but I don't think this is a realistic PR to merge, I just wanted to bring this to your attention.

1. 'Server sent unexpected packet WS.MsgType.CLOSE data 0, aborting'

I would get this when the server I am communicating with would send a close packet with an int 0 data value (failing the if not isinstance(p.data, (str, bytes)) check).  

My hack checks for these 3 types of close or closing or closed packets, constructs a Packet and goes about the self._receive_packet and closing of the write loop.  It seems to work for my use-case, but I am not nearly well versed enough with the engineio internals to know what side-effects this might have (or if it is even a decent solution).

2. 'Packet queue is empty, aborting'

My client is only receiving messages from the server and it seems like on some occasions the server wouldn't send the PING msg in a timely manner, the timeout set on line 580 would expire, and my write queue would be empty.  I chose a somewhat arbitrary 60 seconds of additional padding, but if this is a decent idea it could be a parameter sent when instantiating the AsyncClient?  

As a side question, In my code I also added a concurrent loop running forever just sending a message with data of "PING" to the server every 15 seconds just to keep the write queue populated, is there a better way for a client to send a ping request to a server like this?

Thanks for all your help, feel free to discard this PR, just wanted to bring these issues to your attention and didn't know a better way to do so and share code.

### DIFF
--- a/src/engineio/asyncio_client.py
+++ b/src/engineio/asyncio_client.py
@@ -581,9 +581,9 @@ class AsyncClient(client.Client):
         """
         while self.state == 'connected':
             # to simplify the timeout handling, use the maximum of the
-            # ping interval and ping timeout as timeout, with an extra 25
+            # ping interval and ping timeout as timeout, with an extra 60
             # seconds grace period
-            timeout = max(self.ping_interval, self.ping_timeout) + 25
+            timeout = max(self.ping_interval, self.ping_timeout) + 60
             packets = None
             try:
                 packets = [await asyncio.wait_for(self.queue.get(), timeout)]
--- a/src/engineio/asyncio_client.py
+++ b/src/engineio/asyncio_client.py
@@ -526,7 +526,7 @@ class AsyncClient(client.Client):
                 p = await asyncio.wait_for(
                     self.ws.receive(),
                     timeout=self.ping_interval + self.ping_timeout)
-                if not isinstance(p.data, (str, bytes)):  # pragma: no cover
+                if not isinstance(p.data, (str, bytes, int)):  # pragma: no cover
                     self.logger.warning(
                         'Server sent unexpected packet %s data %s, aborting',
                         str(p.type), str(p.data))
--- a/src/engineio/asyncio_client.py
+++ b/src/engineio/asyncio_client.py
@@ -575,9 +575,9 @@ class AsyncClient(client.Client):
         """
         while self.state == 'connected':
             # to simplify the timeout handling, use the maximum of the
-            # ping interval and ping timeout as timeout, with an extra 5
+            # ping interval and ping timeout as timeout, with an extra 25
             # seconds grace period
-            timeout = max(self.ping_interval, self.ping_timeout) + 5
+            timeout = max(self.ping_interval, self.ping_timeout) + 25
             packets = None
             try:
                 packets = [await asyncio.wait_for(self.queue.get(), timeout)]
--- a/src/engineio/asyncio_client.py
+++ b/src/engineio/asyncio_client.py
@@ -526,7 +526,9 @@ class AsyncClient(client.Client):
                 p = await asyncio.wait_for(
                     self.ws.receive(),
                     timeout=self.ping_interval + self.ping_timeout)
-                if not isinstance(p.data, (str, bytes, int)):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
First of all, putting your changes in a PR is actually very helpful as a way to center the discussion, so I have no problem with that, even if we end up not merging it.

Okay, on the first change, it is really unexpected (as the error says) for the server to send a WebSocket close packet without first closing the Socket.IO and Engine.IO connections. It seems to me this indicates a problem with your server, which is not following the protocol correctly. Is this with my Python server or some other server? In any case, outside of getting the "unexpected" message in your logs, is there any issue with how the connection is aborted? Why do you need to create a fake Engine.IO close packet and feed it through the system?

For the second change you also seem to be trying to compensate for a server problem. If the server is unable to keep up with the ping interval it chose, that seems to indicate a server problem that should be fixed instead of an arbitrary timing extension on the client side.

wescleveland:
From my understanding they are using the nodejs socketio server version 4.5.4 and underlying engineio version 4, the url is blah.appspot.com if that helps at all.

I think for the first issue it might be occurring when they stop/restart the server after a code change, but I'm really not sure as I have no visibility into what they are doing on their end beyond the version and platform info.  It seems to happen a few times in quick succession sometimes which makes me think it's something related to the server stopping/restarting, maybe after a code change or a bug.

The second part I included just because I had seen some people file issues regarding the packet queue is empty error message and wanted to pass along what was happening in my case.  If a client is read-only and the server seems to sometimes delay the ping/pong exchange by a small bit of time, I wanted the flexibility to ignore that and not disconnect.  In my case, disconnecting and reconnecting would make me lose my place in orderbook queues, so I was deep diving into how to prevent that from happening.  With your knowledge of how the server works, is it ok to just fire off "ping" messages to the server every 15 seconds or so to keep the write queue busy?

But ya overall I agree with what you have said and I see how these changes are specific to my edge cases (and unexpected behavior from a server!).

miguelgrinberg:
> From my understanding they are using the nodejs socketio server version 4.5.4

This is very strange, I would not have expected this to be the Node server. Basically the server has time to send a WS close package before going away, but for some reason fails to send the proper Socket.IO and Engine.IO close packets. Makes no sense.

You haven't answered my question though. Besides the error message in the log, are there any issues with the disconnection that happens as a response of the WebSocket close that the server sends you?

> With your knowledge of how the server works, is it ok to just fire off "ping" messages to the server every 15 seconds or so to keep the write queue busy?

The latest version of the Socket.IO protocol has the server sending the pings and the client responding with the pong. The python-socketio 5.x releases respond to pings sent by the server, but have no ping sending ability since the protocol does not require it. If the Node server understand older versions of the Socket.IO protocol, you may want to try connecting with a 4.x python-socketio client, which uses the previous version of the protocol, in which the client sends the ping and the server responds with the pong.

I would be okay adding an option (disabled by default) to ignore ping timeouts in the client. I think it is not a good idea to use it, but if you are dealing with a bad server maybe that helps.

wescleveland:
> You haven't answered my question though. Besides the error message in the log, are there any issues with the disconnection that happens as a response of the WebSocket close that the server sends you?

Sorry missed this, the client handled the disconnect correctly.  I was just digging into why the client was disconnecting and reconnecting so often and encountered that before I addressed the packet queue is empty issue.

The close message might be related to appspot killing connections once an hour, so maybe it's outside the node app and out of their socketio server's control.

Overall, I think with the every 15 seconds ping hack on my end I can deal with the server's funky behavior and get off my branch and back onto your main branch.  I'm hesitant to completely disable the ping timeout for the write queue because I sleep better at night knowing there is one additional thing that might alert me to a dead connection.  I'll leave it up to you to add or ignore, thanks for your time and help on this and all your work on socketio/engineio!

miguelgrinberg:
I'm not sure any of the changes here are still necessary. Please reopen if you are still having trouble with this.

Issue #260, opened 2021-10-24
Apply max_http_buffer_size constant to WebSocket messages


miguelgrinberg:
Current versions of the JS reference implementation use this setting to control the max size for both HTTP and WebSocket incoming packets. This package is currently only using it for HTTP.

PR #310, opened 2023-02-23
Fix documentation for max_http_buffer_size

lawrenceong:
Due to #260, `max_http_buffer_size` is no longer just for polling transport.  Update documentation to reflect change.


miguelgrinberg:
Good catch, thank you!

PR #339, opened 2023-12-29
Not erase sslopt when not ssl_verify

sfonteneau:


miguelgrinberg:
Merged, after I added the missing unit test. Thanks!

PR #345, opened 2024-02-06
Strip root_path to simplify mounting as ASGI sub app

rodja:
As an outcome of https://github.com/encode/starlette/discussions/2413 with newer Starlette versions we now have to provide the full path as `engineio_path` as done in the following example (via `socketio.ASGIApp(socketio_server=sio, socketio_path='/ws/socket.io')`:

```py
# pip install "starlette>=0.33.0" python-socketio uvicorn
import socketio
from starlette.applications import Starlette
from starlette.responses import HTMLResponse
from starlette.routing import Route

async def homepage(_):
    return HTMLResponse("""
    <!DOCTYPE html>
    <html>
    <head>
        <title>SocketIO Test</title>
        <script src="https://cdn.socket.io/4.7.3/socket.io.min.js"></script>
        <script type="text/javascript">
            document.addEventListener('DOMContentLoaded', (event) => {
                var basePath = window.location.pathname;
                var socket = io({ path: basePath + "ws/socket.io" });
                socket.on('msg', function(msg) {
                    document.body.innerHTML = '<h1>' + msg + '</h1>';
                });
            });
        </script>
    </head>
    <body>
        <h1>Connecting to SocketIO...</h1>
    </body>
    </html>
    """)

sio = socketio.AsyncServer(async_mode='asgi', cors_allowed_origins='*')
sio_app = socketio.ASGIApp(socketio_server=sio, socketio_path='/ws/socket.io')
app = Starlette(routes=[Route("/", homepage)])
app.mount('/ws', sio_app)

@sio.on('connect')
async def handle_connect(sid, *args, **kwargs):
    await sio.emit('msg', 'Hello from Starlette!')

# Run the server with "uvicorn demo:app --reload" if your file is called demo.py
```

That works as long as the app is not itself mounted as a sub app like this:

```py
sio = socketio.AsyncServer(async_mode='asgi', cors_allowed_origins='*')
sio_app = socketio.ASGIApp(socketio_server=sio, socketio_path='/ws/socket.io')
sub_app = Starlette(routes=[Route("/", homepage)])
sub_app.mount('/ws', sio_app)
app = Starlette()
app.mount('/sub', sub_app)
```

The above code fails because the frontend tries to connect to `/sub/ws/socket.io` but engineio simply checks if the `scope['path']` starts with `self.engineio_path`.

This pull request fixes this by making python-engineio aware of ASGIs `scope['root_path']`.  It strips the root path from `scope['path']` before checking whether it matches `self.engineio_path`. This allows us to provide the relative `socketio_path` (which is then passed to engineio as `engineio_path`:

```py
sio = socketio.AsyncServer(async_mode='asgi', cors_allowed_origins='*')
sio_app = socketio.ASGIApp(socketio_server=sio, socketio_path='socket.io')
sub_app = Starlette(routes=[Route("/", homepage)])
sub_app.mount('/ws', sio_app)
app = Starlette()
app.mount('/sub', sub_app)
```

The issue came up in https://github.com/zauberzeug/nicegui/issues/2515 and https://github.com/zauberzeug/nicegui/issues/2468. It should be fixed in python-enginio in my opinion to better support "ASGI sub app mounting" for everyone using python-engineio.

### DIFF
--- a/src/engineio/async_drivers/asgi.py
+++ b/src/engineio/async_drivers/asgi.py
@@ -53,11 +53,14 @@ class ASGIApp:
         self.on_shutdown = on_shutdown
 
     async def __call__(self, scope, receive, send):
-        if scope['type'] in ['http', 'websocket'] and \
-                scope['path'].startswith(self.engineio_path):
+        path = scope['path']
+        if 'root_path' in scope and scope['path'].startswith(scope['root_path']):
+            path = scope['path'][len(scope['root_path']):]
+
+        if scope['type'] in ['http', 'websocket'] and path.startswith(self.engineio_path):
             await self.engineio_server.handle_request(scope, receive, send)
         else:
-            static_file = get_static_file(scope['path'], self.static_files) \
+            static_file = get_static_file(path, self.static_files) \
                 if scope['type'] == 'http' and self.static_files else None
             if scope['type'] == 'lifespan':
                 await self.lifespan(scope, receive, send)
--- a/src/engineio/async_drivers/asgi.py
+++ b/src/engineio/async_drivers/asgi.py
@@ -17,9 +17,17 @@ class ASGIApp:
     :param static_files: A dictionary with static file mapping rules. See the
                          documentation for details on this argument.
     :param other_asgi_app: A separate ASGI app that receives all other traffic.
-    :param engineio_path: The endpoint where the Engine.IO application should
-                          be installed. The default value is appropriate for
-                          most cases.
+    :param engineio_path: The endpoint prefix for the Engine.IO application, 
+                          used to distinguish Engine.IO traffic from other 
+                          traffic based on the request path. The default value
+                          suits most scenarios where Engine.IO operates on a 
+                          specific subpath.
+                          If set to None, the middleware assumes all incoming 
+                          traffic is directed to the Engine.IO server, bypassing
+                          path-based filtering. This setting is useful for 
+                          encapsulated ASGI applications. The other_asgi_app and
+                          static_files parameters are ignored when engineio_path
+                          is None.
     :param on_startup: function to be called on application startup; can be
                        coroutine
     :param on_shutdown: function to be called on application shutdown; can be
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
I don't think this is the right change for this package.

The idea of `socketio_path` is that it is the same as `path` in the client. If your client has to use `/sub/ws/socket.io`, then this is what you should set as the Socket.IO path in the server. And then everything works, right? The fact that you mounting apps on apps to build your server is not something the Socket.IO layer needs to care about.

rodja:
Thanks for reviewing my PR and for your insights on the `socketio_path` and ASGIApp dispatcher setup.

I came across a discussion in the [ASGI spec](https://github.com/django/asgiref/issues/229#issuecomment-765583185) that elaborates on `root_path`'s role, indicating internal handling for URL routing is expected from ASGI servers. My intention with this PR is to bring python-engineio in line with this aspect of the ASGI spec, aiming to enhance its adaptability and flexibility in modular setups like [NiceGUI](https://nicegui.io), where encapsulation and self-contained packaging are key.

I realize now that this adjustment could introduce a breaking change, for which I should have initially marked the PR as a draft and highlighted the potential impact more clearly. I'm eager to discuss the path towards ASGI spec compatibility and whether it's a direction you'd consider exploring together, ensuring we provide thorough documentation for a smooth transition.

Looking forward to your feedback on pursuing this alignment with the ASGI spec.

rodja:
> Thinking over this, maybe the best option is to add an option that implements what I just said. For example, setting socketio_path=None disables any path checking and assumes all traffic is Socket.IO traffic. Then you can implementing your routing fully using Starlette.

That is a wonderful idea. I just altered this pull request accordingly. I thought its most consistent if the parameters `static_files` and `other_asgi_app` are ignored when `engineio_path` is set to none because they only make sense when using it as a top-level router. 

An alternative could be to split the two variants into separate classes. For example `AsgiRootApp` and `AsgiSubApp`. What do you think?

PR #361, opened 2024-07-29
Add automatic path handling to ASGIApp

imcarlosguerrero:
Hello! It's me again, I took advice of what you said at #360 

> The default behavior should be that you provide your endpoint with or without trailing slash, and the server accepts the endpoint with or without trailing slash. This should work in pretty much all cases without any need to add configuration. I think just doing this would be an improvement.

The code I propose achieves the following:

With the `automatic_path_handling` parameter as **True**

`foo` results in `/foo`

`foo/foobar` results in `/foo/foobar`

`///foo` results in  `/foo`

`///foo///` results in `/foo`

`////foo/////foobar///` results in `/foo/foobar`

`''` (empty string) results in `/`

`/` results in `/`

With the automatic_path_handling as **False**

`foo` results in `/foo`

`foo/` results in `/foo/`

`/foo/foobar/` results in `/foo/foobar/`

`//foo///` results in `/foo///`

`///foo///foobar////////` results in `/foo///foobar////////`

`''` (empty string) results in `/`

`/` results in `/`

I'm still not sure on whether removing the extra slashes such as the browsers do is right or not, because at least Postman can connect with those weird paths.

Other thing is that even with `automatic_path_handling` as **False**, the leading slashes are removed because a path with multiple leading slashes doesn't make much sense.

I think you can enlighten me on what to do exactly when the option is set to **False**, this pull requests aims more for advice than for a merge right now.

Having in mind that:

> When you remove the trailing slash path matching using startswith() doesn't work anymore.

I changed the behaviour of `__call__`  to trim the trailing slash when the ASGI middleware is called and the `automatic_path_handling` parameter is set to True (which should be the default value).

This way we can match the pattern using `scope['path'] == self.engineio_path` instead of `scope['path'].startswith(self.engineio_path)`, this allows the routes to work with or without the trailing slash and without other possible paths interfering, for example

`/foo` and `/foo/` both work, but `/foobar` doesn't.

If you think this approach could be beneficial, I'd be happy to further refine and extend it to the WSGIApp as well. Your feedback would be greatly appreciated, as it will help me determine whether this change aligns with the project goals. If it’s not something that's being considered at this time, please let me know, and I'll focus my efforts elsewhere and just use a custom wheel in my projects to have this functionality.

**Note**: This change is useful in the case I described at #359 because I couldn't get an ASGI middleware to work to trim the trailing slash in the `socketio_app` while keeping the `other_asgi_app` working correctly.

### DIFF
--- a/src/engineio/async_drivers/asgi.py
+++ b/src/engineio/async_drivers/asgi.py
@@ -1,9 +1,9 @@
-import re
 import os
 import sys
 import asyncio
 
 from engineio.static_files import get_static_file
+
 
 class ASGIApp:
     """ASGI application middleware for Engine.IO.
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Here is some feedback:

- There is no need to have a setting. The `True` setting is what we want, the `False` setting is unnecessary as far as I can see.
- There is no need to remove duplicate slashes. Garbage in, garbage out, as they say.
- The Socket.IO path should be stored with leading and trailing slashes. If the user does not include them, they should be added in the constructor.
- The incoming request path should always have a trailing slash. If it is missing in the request, then it should be added
- The path comparison should continue to use `startswith()`. Changing this could break some use cases.

imcarlosguerrero:
Hello, I ended up just adding the trailing slash to the incoming request if it is missing. I tested it and everything seems to work fine with or without the trailing slash being included in the request, let me know what you think about this, please.

imcarlosguerrero:
This code breaks the `other_asgi_app`, sorry, I'll fix it.

imcarlosguerrero:
@miguelgrinberg I added the trailing slash to the incoming request but that generates a problem with the `other_asgi_app`. It creates an infinite loop of redirects.

`[0] INFO:     127.0.0.1:16292 - "GET /api/ HTTP/1.1" 307 Temporary Redirect`
`[0] INFO:     127.0.0.1:16295 - "GET /api/ HTTP/1.1" 307 Temporary Redirect`
`[0] INFO:     127.0.0.1:16295 - "GET /api/ HTTP/1.1" 307 Temporary Redirect`
`[0] INFO:     127.0.0.1:16295 - "GET /api/ HTTP/1.1" 307 Temporary Redirect`

What I did to solve it was to add a check where if the incoming request on the case that it is not going the way of the Socket.IO path, then remove that added trailing slash, but I don't know if that's the right way to go, what do you think?

miguelgrinberg:
I have committed a simpler solution now. Thanks!

PR #387, opened 2024-11-25
Replace asyncio.iscoroutinefunction with inspect.iscoroutinefunction

musicinmybrain:
Fixes `DeprecationWarning` on Python 3.14, although for now we still get them from `pytest-asyncio`; `asyncio.iscoroutinefunction` is scheduled for removal in Python 3.16.

https://docs.python.org/3.13/library/inspect.html#inspect.iscoroutinefunction


miguelgrinberg:
Considering that there's still 3 more versions until removal I'm going to keep things the way they are for now and apply this change once I have phased out support for 3.8 and 3.9.

musicinmybrain:
> Considering that there's still 3 more versions until removal I'm going to keep things the way they are for now and apply this change once I have phased out support for 3.8 and 3.9.

Sounds reasonable to me. The [documentation for `inspect.iscoroutinefunction`](https://docs.python.org/3.13/library/inspect.html#inspect.iscoroutinefunction) didn’t give me any reasons to expect this not to work in 3.8 and 3.9, but the CI certainly seems to be showing that there is some kind of difference in behavior between `inspect.iscoroutinefunction` and `asyncio.iscoroutinefunction` on those Python versions.

miguelgrinberg:
The failures are caused, I think, by `AsyncMock` instances.

```python
❯ python3.8
Python 3.8.20 (default, Nov 17 2024, 10:12:25)
Type "help", "copyright", "credits" or "license" for more information.
>>> from unittest.mock import AsyncMock
>>> import inspect
>>> x = AsyncMock()
>>> inspect.iscoroutinefunction(x)
False

❯ python3.13
Python 3.13.0 (main, Nov 17 2024, 23:19:47) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from unittest.mock import AsyncMock
>>> import inspect
>>> x = AsyncMock()
>>> inspect.iscoroutinefunction(x)
True
```

Issue #389, opened 2024-12-06
Decoding unexpected header format causes errors


duckboy81:
**Describe the bug**
When a header contains an unexpected character (e.g. a non-breaking space "\xa0"), the request errors out, preventing further connection.

**To Reproduce**
Execute a `GET` request like below where one of the headers contains a non-breaking space
```js
await fetch("<YOUR HOST HERE>/socket.io/?EIO=4&transport=polling&t=1234567", {
  "headers": {
    "accept": "*/*",
    "x-random-header": `non-breaking-space: \u00A0`
  },
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": null,
  "method": "GET",
  "mode": "cors",
  "credentials": "include"
});
```

**Expected behavior**
I would expect the server to continue to serve the request without issue.

**Logs**
```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "C:\duckboy81\niprgpt\chainlit_client\chainlit\venv\Lib\site-packages\uvicorn\protocols\http\httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\duckboy81\niprgpt\chainlit_client\chainlit\venv\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\duckboy81\niprgpt\chainlit_client\chainlit\venv\Lib\site-packages\engineio\async_drivers\asgi.py", line 67, in __call__
    await self.engineio_server.handle_request(scope, receive, send)
  File "C:\duckboy81\niprgpt\chainlit_client\chainlit\venv\Lib\site-packages\socketio\async_server.py", line 451, in handle_request
    return await self.eio.handle_request(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\duckboy81\niprgpt\chainlit_client\chainlit\venv\Lib\site-packages\engineio\async_server.py", line 209, in handle_request
    environ = await translate_request(*args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #390, opened 2024-12-06
Safer header handling

duckboy81:
Re #389 

Updated the header value decoding to handle unexpected characters. Likely not necessary to update header name decoding due to RFC7230 requirements on improper header names.

### DIFF
--- a/src/engineio/async_drivers/asgi.py
+++ b/src/engineio/async_drivers/asgi.py
@@ -190,7 +190,7 @@ async def translate_request(scope, receive, send):
 
     for hdr_name, hdr_value in scope['headers']:
         hdr_name = hdr_name.upper().decode('utf-8')
-        hdr_value = hdr_value.decode('utf-8')
+        hdr_value = hdr_value.decode('utf-8', errors='replace')
         if hdr_name == 'CONTENT-TYPE':
             environ['CONTENT_TYPE'] = hdr_value
             continue


miguelgrinberg:
The problem with this solution is that it provides an incorrect value for the header. Rejecting the request (possibly without raising an exception) seems to me like a better alternative if we assume that the request is malformed.

As far as I remember the specification says that header values have to be ASCII encoded, which means that UTF-8 should have no problem decoding them. What client is sending these headers that cannot be decoded? Maybe the problem is in the client.

duckboy81:
After doing some more research, I think you're mostly right.



[RFC Section 5.5](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values]) says

> Field values are usually constrained to the range of US-ASCII characters [USASCII]. Fields needing a greater range of characters can use an encoding, such as the one defined in [RFC8187]. Historically, HTTP allowed field content with text in the ISO-8859-1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding. Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB. A recipient SHOULD treat other allowed octets in field content (i.e., obs-text) as opaque data.

Suggests that the header value is not absolutely strict, though using `obs-text` raw is discouraged.

Maybe a happy middle is to wrap that decode call in a try block and at least gracefully handle the error and return a 400. Ideally this would include an error message to the console with what the issue is--this would be a nightmare to hunt down as a dev if all they saw was the 400.

Thoughts?

miguelgrinberg:
Yes, returning a 400 makes sense. You can also use the same logic for the query string, which is also decoded from bytes.

miguelgrinberg:
Bad headers or query strings are now omitted. Implemented in 44ce778c9a7e26d62254563e3de8c4a0c073bdc0.

PR #398, opened 2025-04-04
Pass additional args to callable option cors_allowed_origins

wft-swas:
If option cors_allowed_origins is callable, pass additional args for default_origins and environ to it. Use try-except to still allow usage of a callback without these additional args.

Background:
For our project, we need to dynamically check for origins, which are allowed _additionally_ to the default ones (especially the same origin).

The try-except is just a proposal to avoid a breaking change.


miguelgrinberg:
Thanks! I have simplified the CORS allowed hosts calculation a bit and changed your proposal call signature to include only `origin` and `environ`.

Issue #403, opened 2025-05-03
Incompatibility with gevent 25.4.1


cyfky:
**Describe the bug**
gevent 25.4.1 (more precisely, gevent/gevent@25a53bf782c41470d94ef209a4413180561c16c4) added `Queue` to the monkey patch list, causing the following line to throw an error:
https://github.com/miguelgrinberg/python-engineio/blob/3eafe26d47fabd0f03853b29e0e7c27cc78a66b5/src/engineio/client.py#L166

**To Reproduce**
1. Run `pip install --upgrade gevent==25.4.1 "python-engineio[client]"`
2. Create a file `main.py` with the following content:
```
import gevent.monkey
gevent.monkey.patch_all()
import engineio

eio = engineio.Client()
q = eio.create_queue()
print(q)
```
3. Run `python main.py`, and you get the following error:
```
AttributeError: 'gevent._gevent_cqueue.Queue' object has no attribute 'Empty' and no __dict__ for setting new attributes. Did you mean: 'empty'?
```

**Expected behavior**
If you change the first command to `pip install --upgrade gevent==24.11.1 "python-engineio[client]"` you get the correct output:
```
<queue.Queue object at (memory_addr)>
```

**Additional context**
[A similar problem](https://github.com/gevent/gevent/issues/2099) caused by this change has been reported; as a response, a flag has been added in gevent 25.4.2 via gevent/gevent@926b0d3c2ed09bfa994689419dd75723d3bc6b00. Indeed, running
```
pip install --upgrade gevent==25.4.2 "python-engineio[client]"
GEVENT_MONKEY_DISABLE_QUEUE_QUEUE=true python main.py
```
gives the correct output.

cyberw:
@miguelgrinberg Hi Miguel! 

Do you think you'll get the chance to take a look at this any time soon? Latest gevent has a few security patches we'd very much like. No idea if it is an easy fix or not...

miguelgrinberg:
I haven't looked at this yet, but see the comment above yours for the workaround. You just need to set an environment variable to tell gevent to revert the Queue to the old behavior.

cyberw:
Oh, I thought the ”workaround” completely disabled monkey patching of Queue?

cyberw:
Thanks for the explanation, then I can use the workaround for now! Great!

PR #409, opened 2025-06-02
Fixing miguelgrinberg/python-engineio#403.

cyfky:
# Fixes
This is a minimal fix for #403.

# Changes
1. Storing the queue empty exception in the client instance, so that we no longer require the ability to add attributes to queue.Queue. This solves the problem introduced by gevent 25.4.1, where the monkey-patched queue.Queue becomes Cython-based and no longer allows adding attributes.
2. Introducing a class `MockQueue` in `tests/common/test_socket.py` as the queue provider for the mock server, so that later tests that modify its `.join` method can run.
3. Adding a new test environment for checking compatibility with different versions of gevent. This can be invoked as `tox -e gevent`. This will first monkey-patch everything, then run the tests described in the main `[testenv]` environment, with the exception that all tests inside `tests/common/test_server.py` having `gevent` in their names will be skipped (as these tests interfere with monkey-patching). An optional env var called `GEVENT_VERSION` (defaults to "25.4.1") can be passed in to specify the version of gevent to be used.

# Tests Done
1. `tox -e py` (default testing, without monkey patching) and `tox -e gevent` both pass.
`GEVENT_VERSION=24.11.1 tox -e gevent` also passes.
2. Performance tests (`tests/performance/run.sh`) show no noticeable difference. 
Results from my local runs are as follows:
> 
> Before patching:
> 
>> text_packet: 2022430 packets processed.
>> binary_packet: 16274018 packets processed.
>> binary_b64_packet: 9206001 packets processed.
>> json_packet: 1840093 packets processed.
>> payload: 867860 payloads processed.
>> server_receive: 969609 packets received.
> 
> After patching:
> 
>> text_packet: 2106527 packets processed.
>> binary_packet: 16469943 packets processed.
>> binary_b64_packet: 9495183 packets processed.
>> json_packet: 1817357 packets processed.
>> payload: 851590 payloads processed.
>> server_receive: 960605 packets received.

### DIFF
--- a/src/engineio/async_client.py
+++ b/src/engineio/async_client.py
@@ -202,7 +202,7 @@ class AsyncClient(base_client.BaseClient):
     def create_queue(self):
         """Create a queue object."""
         q = asyncio.Queue()
-        q.Empty = asyncio.QueueEmpty
+        self.queue_empty_exc = asyncio.QueueEmpty
         return q
 
     def create_event(self):
@@ -624,7 +624,7 @@ class AsyncClient(base_client.BaseClient):
             packets = None
             try:
                 packets = [await asyncio.wait_for(self.queue.get(), timeout)]
-            except (self.queue.Empty, asyncio.TimeoutError):
+            except (self.queue_empty_exc, asyncio.TimeoutError):
                 self.logger.error('packet queue is empty, aborting')
                 break
             except asyncio.CancelledError:  # pragma: no cover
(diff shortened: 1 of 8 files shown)


miguelgrinberg:
Thanks for your PR, but I have decided to take a different approach that is less convoluted. This issue is now fixed in dcde006b62e004c8a1d29d3478c95fbffbb11122.

PR #410, opened 2025-06-12
Support set cors allowed origins as `['*']`

wu-clan:
Hi, @miguelgrinberg 

This change will be compatible with the following:
https://github.com/encode/starlette/blob/78da9b9e218ab289117df7d62aee200ed4c59617/starlette/middleware/cors.py#L19

The same will apply to fastapi, which helps synchronize the setup of server and ws cors allowed origins.


miguelgrinberg:
Why is this needed?

PR #415, opened 2025-09-26
Reset client queue upon disconnection (Fixes #414)

### LINKED ISSUE #414: A client object that is reused may have left over data in the packet queue
The issue is described in a [discussion](https://github.com/miguelgrinberg/Flask-SocketIO/discussions/2144#discussioncomment-14519753).

Basically the problem appears to occur with certain timings of the reconnect feature of the python-socketio `Client` and `AsyncClient` classes, which reuse the Engine.IO object. The queue that stores outgoing packets sometimes ends up with data when a connection breaks in certain specific ways, and this data is not cleaned, so it is pulled by the reconnected client, which gets all confused and, at least in the case of the issue referenced above, disconnects immediately after reconnecting.

miguelgrinberg:


Joe-ChenZ:
This seems to work for me in my application. What would be the ETA of releasing this publicly?

miguelgrinberg:
A day or two. I need to rework some unit tests to adapt them to this change first.

PR #417, opened 2025-10-17
Fix Cookie Parsing in WebSocket Connections

meshya:
Engine.IO clients fail to handle quoted cookie values when upgrading to WebSocket connections, causing connection issues with applications that use cookies containing spaces or special characters.

### Specific Issues:

`AsyncClient`: 
1. Manual string splitting crashes on name="Alice; Bob"
2. Misbehavior in some cases.
```python
input_cookie_string = 'c1=v1; c2="v2"'
# This cookie will parses like:
output_cookie_dict = {
    'c1': 'v1',
    'c2': '"v2"'
}
# Correct form is:
output_cookie_dict = {
    'c1': 'v1',
    'c2': 'v2'
}
``` 
`Sync client`:
1. Inconsistent cookie encoding between HTTP and WebSocket transports
2. Misbehavior in some cases.
```python
cookie_dict = {
    'foo': 'bar',
    'fooo': '"bar"'
}

# If we pass it to http_session it will take effect in cookie header like this
cookie_header = 'foo=bar; fooo="bar"'

# But correct form is:
cookie_header = 'foo=bar; fooo=\\"bar\\"'

```
### Solution:

Replace manual cookie parsing with Python's standard `http.cookies.SimpleCookie`

1. Added client_cookies.py utility module
2. Fixed async client cookie header parsing
3. Fixed sync client cookie encoding consistency
4. Updated test cases to reflect correct behavior

I’m open to feedback and ready to address any issues or improvements in this PR.

### DIFF
--- a/src/engineio/async_client.py
+++ b/src/engineio/async_client.py
@@ -12,6 +12,7 @@ from . import base_client
 from . import exceptions
 from . import packet
 from . import payload
+from . import client_cookies
 
 async_signal_handler_set = False
 
@@ -319,14 +320,12 @@ class AsyncClient(base_client.BaseClient):
 
         # extract any new cookies passed in a header so that they can also be
         # sent the the WebSocket route
-        cookies = {}
         for header, value in headers.items():
             if header.lower() == 'cookie':
-                cookies = dict(
-                    [cookie.split('=', 1) for cookie in value.split('; ')])
+                cookie_dict = client_cookies.decode_cookie_string(value)
+                self.http.cookie_jar.update_cookies(cookie_dict)
                 del headers[header]
                 break
-        self.http.cookie_jar.update_cookies(cookies)
 
         extra_options = {'timeout': self.request_timeout}
         if not self.ssl_verify:
--- a/src/engineio/client.py
+++ b/src/engineio/client.py
@@ -19,6 +19,7 @@ from . import base_client
 from . import exceptions
 from . import packet
 from . import payload
+from . import client_cookies
 
 default_logger = logging.getLogger('engineio.client')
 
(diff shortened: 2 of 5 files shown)


meshya:
Glad to hear you liked the main idea!

Regarding the test changes — it’s not that I *modified them just to make everything pass*. The issue actually comes from an inconsistent behavior in the `_connect_websocket` function of the `Client` module, which the tests currently depend on.

When you create a `Client` with an `http_session`, like this:
```python
from engineio.client import Client
import requests

http = requests.Session()
c = Client(http_session=http)
c.connect('ws://foo', transports=['websocket'])
```
the call to `websocket.create_connection` receives `cookie=''`.
But if no `http_session` is passed, it receives `cookie=None`.

This difference originates from the logic in `client.py` (around lines 267–279):
```python
        cookies = None
        extra_options = {}
        if self.http:  # passing http_session makes a difference even if http.cookies is empty
            cookies = '; '.join([f"{cookie.name}={cookie.value}"
                                 for cookie in self.http.cookies])
            for header, value in headers.items():
                if header.lower() == 'cookie':
                    if cookies:
                        cookies += '; '
                    cookies += value
                    del headers[header]
                    break
```

As a result, tests like  
- `test_websocket_connection_with_auth`  
- `test_websocket_connection_with_cert`  
- `test_websocket_connection_with_cert_and_key`  
- `test_websocket_connection_verify_with_cert_and_key`  
- `test_websocket_connection_with_proxies`  
- `test_websocket_connection_without_verify`  
- `test_websocket_connection_with_verify`  

expect `cookie=''`, while others such as  
- `test_polling_connection_not_upgraded`  
- `test_websocket_connection_extra`  
- `test_websocket_connection_failed_with_websocket_error`  
- `test_websocket_upgrade_failed`  

expect `cookie=None`.

So the test updates were meant to reflect that inconsistent behavior, not to “force” tests to pass.

I didn’t use any AI assistance for this PR — I found the issue, read through the client code carefully, and tried to propose a solution that fits your style. I only use AI tools to help me write messages, not to create or modify code.

I’d really appreciate another chance to improve the implementation in a simpler way that aligns better with your expectations. Contributing to `python-engineio` is a great learning opportunity for me, and I’m eager to refine this further based on your feedback.

PR #420, opened 2025-10-30
[RFC] Remove logger configuration from client/server classes

ZipFile:
I was playing around with socketio/engineio and stumbled upon dubious behavior. Some of my logs got duplicated! Upon investigation, I've found that engineio adds `StreamHandler` to the default logger. Which is unexpected, to say the least. Documentation was also misleading, since passing `logger=False` does not disable this behavior, just sets different log level.

Note: I'm observing this behavior because I'm configuring loggers after instantiating the app, but before running it. Questionable architecture decision for sure, but I do not think I'm alone in this regard.

So, my proposal is following:
* Do not mess with loggers, accept whatever user passed or fallback to default. It's not the library responsibility to configure loggers after all.
* Remove misleading documentation section. Since there are no additional logic here now, I do not think wording like "if you need more debug-ability - configure logging" would fit.

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -336,27 +336,6 @@ Or for ``asyncio``::
 The single argument passed to the method is the number of seconds to sleep
 for.
 
-Debugging and Troubleshooting
------------------------------
-
-To help you debug issues, the server can be configured to output logs to the
-terminal::
-
-    import engineio
-
-    # standard Python
-    eio = engineio.Server(logger=True)
-
-    # asyncio
-    eio = engineio.AsyncServer(logger=True)
-
-The ``logger`` argument can be set to ``True`` to output logs to ``stderr``, or
-to an object compatible with Python's ``logging`` package where the logs should
-be emitted to. A value of ``False`` disables logging.
-
-Logging can help identify the cause of connection problems, 400 responses,
-bad performance and other issues.
-
 .. _deployment-strategies:
 
 Deployment Strategies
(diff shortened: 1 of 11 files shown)


miguelgrinberg:
No, I'm sorry, this is wrong. If you want to use your own logger, then pass it as an argument, and it will not be touched.

ZipFile:
> If you want to use your own logger, then pass it as an argument, and it will not be touched.

Well, that's why I've created this PR in the first place. Code like this:

```python
from engineio.server import Server
from engineio.base_server import default_logger
s = Server(logger=default_logger)
```

looks super wrong to me. I do not want to create any new loggers, just wish nothing would touch existing ones.

Maybe defaulting `logger` to `default_logger` instead of `False` would be a less wrong solution?

miguelgrinberg:
> That's pretty much what I did.

That is not what you did. You imported a private variable used internally by this package and used it in your application. My example does not do that and I do not see anything awkward about it.

ZipFile:
> What we disagree is what should happen when the application does not pass a logger.

Yep, my point is that current behavior is not obvious (library code typically never do so), under-documented (logger ownership) and can get you easily into the situation I've got (double logging). I want this to change.

> And I think we are not going to come to an agreement there.

Well, let be it. All my past experience with Python logging tells me current behavior is wrong, yours might be different. Either way, I've raised the issue and now it should be at least searchable on the internet.

Before we cease the discussion altogether, should I create proper issue (based on https://github.com/miguelgrinberg/python-engineio/pull/420#issuecomment-3468006180)?

PR #448, opened 2026-06-09
Fire disconnect event when write loop times out on empty queue

Obi-Dann:
## Motivation

We hit this in production with the threading websocket client: it was left permanently in the `connected` state with no `on_disconnect` ever firing, even though the underlying connection was dead. Because the disconnect event never fired, application-level reconnect logic never triggered, and the client stayed wedged until we recreated it (instantiating a fresh client and reconnecting).

## Problem

When `_write_loop`'s `queue.get(timeout=...)` expires — because the server stopped sending pings and no PONG was queued — the client silently exits the write loop without closing the connection or notifying user code.

The root cause is a timeout asymmetry between the two background loops:

- write loop: `max(ping_interval, ping_timeout) + 5` (40s with defaults)
- websocket read loop `recv()`: `ping_interval + ping_timeout` (60s with defaults)

When the write loop's timeout expires first, `_read_loop_websocket` is still alive — it keeps receiving server pings, each one enqueuing a PONG that will never be sent. The client never fires `on_disconnect` and never transitions out of the `connected` state. This wedge can persist indefinitely against a server that keeps sending occasional traffic but doesn't enforce its own PONG timeout.

In Engine.IO v4 the client only PONGs in response to server pings (it never originates pings), so on an idle connection the write queue's only natural feeder is that PONG. The empty-queue timeout therefore only fires when the server has genuinely stopped pinging for ~40s — i.e. a dead connection — making silent abort the wrong response.

## Fix

In the empty-queue branch of `_write_loop`, wake the read loop in a transport-correct way so its **existing** exit path owns the disconnect (fires the event, removes the client from `connected_clients`, resets state):

- **websocket** → `self.ws.close()`, which unblocks the read loop blocked in `recv()` / `receive()`. This mirrors what `disconnect()` already does.
- **polling** → `self.write_loop_task = None`, which makes `_read_loop_polling`'s `while self.state == 'connected' and self.write_loop_task:` condition fall through — the same mechanism already used by the polling HTTP-error path.

The write loop never fires the disconnect event itself, preserving the codebase's single-owner invariant (the read loop is the sole place that triggers `disconnect` / `_reset()`), so there is no double-fire. Graceful `disconnect()` is unaffected: it enqueues `None` first, so the write loop exits through the normal sentinel path, never this timeout branch.

The same change is applied to `asyncio_client.py`.

## Testing

New tests in both `tests/common/test_client.py` and `tests/async/test_client.py`:

- `test_write_loop_empty_queue_polling` / `..._websocket` — focused unit tests pinning each branch (polling nulls `write_loop_task`; websocket calls `ws.close()` and leaves `write_loop_task` truthy).
- `test_write_loop_empty_queue_propagates_disconnect` — end-to-end test running `_write_loop` and `_read_loop_websocket` concurrently (real threads / asyncio tasks), asserting the client leaves `connected` and fires exactly one `on_disconnect`.

All new tests fail on `main` (state stays `connected`) and pass with the fix. Full suite passes and `tox -eflake8` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/engineio/async_client.py
+++ b/src/engineio/async_client.py
@@ -638,6 +638,15 @@ class AsyncClient(base_client.BaseClient):
                 packets = [await asyncio.wait_for(self.queue.get(), timeout)]
             except (self.queue_empty, asyncio.TimeoutError):
                 self.logger.error('packet queue is empty, aborting')
+                # signal the read loop to exit so it fires the disconnect event
+                if self.current_transport == 'websocket' and \
+                        self.ws is not None:
+                    try:
+                        await self.ws.close()
+                    except Exception as e:
+                        self.logger.warning('Error closing WebSocket: %s', e)
+                else:
+                    self.write_loop_task = None
                 break
             except asyncio.CancelledError:  # pragma: no cover
                 break
--- a/src/engineio/client.py
+++ b/src/engineio/client.py
@@ -584,6 +584,15 @@ class Client(base_client.BaseClient):
                 packets = [self.queue.get(timeout=timeout)]
             except self.queue_empty:
                 self.logger.error('packet queue is empty, aborting')
+                # signal the read loop to exit so it fires the disconnect event
+                if self.current_transport == 'websocket' and \
+                        self.ws is not None:
+                    try:
+                        self.ws.close()
+                    except Exception as e:
+                        self.logger.warning('Error closing WebSocket: %s', e)
+                else:
+                    self.write_loop_task = None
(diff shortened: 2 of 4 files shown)


miguelgrinberg:
Read the contribution guidelines before submitting PRs to my projects please.

Obi-Dann:
Sorry, missed the tab with the contribution guidelines ❤️

